PR #15, opened 2020-06-10
Updates to issue templates and all that exciting stuff

wreality:
There's a few things that likely need a second set of eyes before they get merged in things like:

* Code of Conduct (@kfitz / @sattler9 )
* Contribution guidelines (@Sruthin86 )
* License (@kfitz )

I went with GPLv3 for the license for a couple of reasons.  Manifold is GPLv3 and they're also a Mellon funded project, so why not join the party.  Also, if we were to go with MIT, we wouldn't be able to use GPLv3 libraries (like-license restrictions in the GPL and all).

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -1,0 +1,97 @@
+# How to contribute
+
+MESH-Research loves to welcome your contributions. There are several ways to help out:
+
+* Create an [issue](https://github.com/MESH-Research/CCR/issues) on GitHub, if you have found a bug
+* Write test cases for open bug issues
+* Write patches for open bug/feature issues, preferably with test cases included
+
+There are a few guidelines that we need contributors to follow so that we have a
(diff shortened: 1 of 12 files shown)


.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,97 @@
...
+
+    [optional footer(s)]
+
+Making use of [tooling which supports Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#tooling-for-conventional-commits) is highly encouraged.
+
+## Submitting Changes
+
+* Push your changes to a topic branch in your fork of the repository.
+* Squash extraneous commits to the appropriate logical portions.
+* Submit a pull request to the repository in the CakePHP organization, with the
```
wreality, line 60:
```suggestion
* Submit a pull request to the repository in the MESH-Research Organization, with the
```

PR #28, opened 2020-12-23
Laravel 8 upgrade

wreality:


backend/docker-compose.yml
```diff
@@ -0,0 +1,74 @@
+# For more information: https://laravel.com/docs/sail
```
gmeben, line 1: I was under the impression that Lando would've eliminated the need for this file.

backend/app/Models/User.php
```diff
@@ -0,0 +1,43 @@
...
+{
+    use HasFactory, Notifiable;
+
+    /**
+     * The attributes that are mass assignable.
+     *
+     * @var array
+     */
+    protected $fillable = [
+        'name',
```
gmeben, line 20: This should probably be called `username` right?

backend/docker-compose.yml
wreality: That docker compose comes with the Laravel project. I'm leaving it there for the time being as a reference until we decide how our production containers will be defined.

backend/app/Models/User.php
wreality: Defaults from Laravel. I was going to make these changes in a separate PR. You're right though, I'll update it to match how things worked before the upgrade.

gmeben (approved):
I got this up and running locally without issue. Everything looks good as far as I'm concerned.

PR #31, opened 2021-01-06
fix: update axios dependency due to SSRF vulnerability

wreality:
This upgrades the Axios version and should clear the existing dependabot alert.

Axios can probably be removed in the future as the apollo integration is completed, but removing it now would break the build and create a lot of work to extract it.

We should extract it as part of the apollo-client integration tasks.


wreality (commented):
These are the same changes that would have been implemented by #30 which has been reviewed.

Issue #62, opened 2021-01-07
Setup the connection between Apollo Client and Lighthouse


wreality:
**Problem/Why**

There is currently no way for the frontend and backend to communicate with one another on CCR.

---

**Goal/What**

The frontend and backend applications need a way to communicate with each other. Connecting Apollo Client and Lighthouse alleviates the problem. Ensure successful data transfer between the frontend and backend apps.

---

**Solution/How**

--- 
## Attachments 
 
- [Is child to System User Authentication](https://trello.com/c/XK8oNtaj/11-system-user-authentication#sf-dependencies&dependencyType=is_child_to&dependencyId=5f68d540d839a876b6bc7551&resolved=false) 
 
> Trello Source: https://trello.com/c/kYJzyych

wreality:
There's honestly not much to review here since it seems most of the work had already been completed, but still could @gmeben or @Sruthin86 give PR #89 a look over and mark it ask reviewed please & thanks.

gmeben:
@wreality I approved pull request #89

PR #89, opened 2021-01-08
fix: remove hostname from graphql api config

wreality:
#62


gmeben (approved):
Can't confirm anything beyond **not** encountering any unexpected issues when I run this code locally. As expected, I get an "EOF" error when hitting the GraphQL endpoint.

Issue #60, opened 2021-01-07
Internal user creation


wreality:
**Problem/Why**

Users cannot register an account.

---

**Goal/What**

Users should be able to create a local account (not including social sign-on).

---

**Solution/How**

Implement front-end pages and API endpoints to allow a user to register.  Input data should be validated on both the front-end and backend and appropriate feedback should be provided to users.

** Schema **
```
Query {
   userCreate(name: String!, email: String!, password: String!): User!
}
```
Validation Criteria:

- Name: required
- Email: required, valid email, unique
- Password: required, zxcvbn at least 5 (?)

Password should be validated on both frontend and backend.  Live feedback should be provided to users regarding the strength of their password.

Laravel zxcvbn library: https://github.com/olssonm/laravel-zxcvbn

--- 
### Review Checklist 
- [ ] Users can register an account 
- [ ] Submitted data is validated properly. 
- [ ] UI/Flow follows best-practices and provides for excellent User experience. 
### To Do 
- [x] Configure apollo-client to communicate with the laravel API 
- [ ] Add tests as appropriate 
 
> Trello Source: https://trello.com/c/M3LP21fR

Sruthin86:
The developers worked together on the implementing the GraphQL schema for user creation. Code is checked in and a pull request created. 
MESH-Research/CCR: Pull Request 94

PR #94, opened 2021-01-13
feat: add user creation api

Sruthin86:
Add GraphQL schema for user creation

### DIFF
--- a/backend/app/Models/User.php
+++ b/backend/app/Models/User.php
@@ -50,8 +50,6 @@ class User extends Authenticatable
      */
     public static function createRules() {
         return [
-        'first_name' => 'required|max:55',
-        'last_name' => 'required|max:55',
         'username' => ['required', new Username, 'max:15', 'unique:users'],
         'email' => 'email|required|unique:users',
         'password' => 'required|zxcvbn_min:4|max:255',        
--- a/backend/app/Models/User.php
+++ b/backend/app/Models/User.php
@@ -44,35 +44,6 @@ class User extends Authenticatable
     ];
 
     /**
-     * Rules for creation of a user
-     *
-     * @return array
-     */
-    public static function createRules() {
-        return [
-        'username' => ['required', new Username, 'max:15', 'unique:users'],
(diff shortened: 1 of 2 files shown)


backend/app/Models/User.php
```diff
@@ -50,8 +50,6 @@ class User extends Authenticatable
      */
     public static function createRules() {
```
wreality, line 51: We should remove the createRules and updateRules methods, since they don't really do anything

backend/graphql/schema.graphql
```diff
@@ -1,18 +1,50 @@
-"A date string with format `Y-m-d`, e.g. `2011-05-23`."
+"A date string wit`h format `Y-m-d`, e.g. `2011-05-23`."
```
wreality, line 1:
```suggestion
"A date string with format `Y-m-d`, e.g. `2011-05-23`."
```

backend/graphql/schema.graphql
```diff
@@ -1,18 +1,50 @@
...
+    name: String
+    
+    "Username. Must be unique. (required)"
+    username: String! @rules(apply: ["unique:users,username"],  messages: { unique: "Username is already in use"})
+    
+    "Email address. Must be unique. (required)"
+    email: String! @rules(apply: ["unique:users,email"], messages: { unique: "Email address is already in use"})
+    
+    "Password used to login to the application. (required)"
+    password: String! @hash
```
wreality, line 29: We need to add checking password complexity with https://github.com/olssonm/laravel-zxcvbn

Check out the validation criteria in #60

wreality (changes requested):
We forgot about validating password complexity, but other than that this appears good.  If we add the zxcvbn validation I think this'll be good to merge.

backend/app/Models/User.php
wreality: REmoved.

backend/graphql/schema.graphql
wreality: Added

wreality (approved):

Issue #91, opened 2021-01-08
User creation frontend functionality


wreality:
**Problem/Why**

There is no way to create an account in the application

---

**Goal/What**

A user is able to create an account.  The forms should follow best-practices for user sign-up and provide an excellent user experience.

---

**Solution/How**

Implement frontend functionality to allow a user to create a user account.  Validate user data before sending to the API as much as possible.  Sensible feedback should be provided to users about failures and how to remedy those failures.

** See parent card #60 for schema **
---
### To Do / Acceptance Criteria
- [x] Implement login page(s)
- [x] Implement validation criteria
- [x] Send user input to API
- [x] Provide feedback to the user about the result of their request.
### Review Checklist
- [ ] Excellent UI/UX
- [ ] Able to create new user
- [ ] Sensible feed back on result of request.
- [ ] Testing is implemented for any new code.
### Accessibility Checklist
- [ ] WCAG

wreality:
How to mock schema methods before they're available:

https://www.apollographql.com/docs/react/development-testing/client-schema-mocking/

PR #96, opened 2021-01-17
feat: add ability to create user account from client

wreality:
User creation front end functionality #91.
* Update interface for Register form
* Validate data locally as much as possible before submitting
* Utilize createUser mutation for creating a user
* Unit test new components.

Reference #91 for acceptance criteria & #60


client/src/pages/Login.vue
```diff
@@ -13,11 +13,11 @@
             square
             ref="username"
             v-model="form.username"
-            :label="$t('auth.login_fields.username')"
+            :label="$t('auth.fields.username')"
             @keypress.enter="$refs.password.focus()"
             autofocus
           >
-            <template v-slot:prepend>
+            <template #prepend>
```
gmeben, line 20: Nothing that this is the only place in this file that uses the `v-slot` shorthand.

gmeben (approved):

client/src/pages/Login.vue
wreality: I'll fix the other one.  Admittedly I don't think I intended to include this change, but I do prefer the shorthand (which is probably why it got picked up... I probably did a find and replace on s/v-slot:/#/ )

wreality:
@gmeben I've got a couple of little things that I noticed going back through the I'm going to fix before merging. Thanks for the review!

client/src/pages/Login.vue
gmeben: Yeah, I noticed you changed it to the shorthand elsewhere. Seems like making this stylistic change consistent throughout the codebase could be a separate pull request itself.

PR #97, opened 2021-01-19
feat: Sanctum setup

gmeben:
This pull request:

- Unignores `vendor/` so `/backend/resources/views/vendor/graphql-playground/index.blade.php` can be versioned and because the `/backend/vendor` directory is already ignored
- Adds Laravel Sanctum as a dependency
- Adds Laravel Sanctum middleware to Kernel and Lighthouse
- Adds Lighthouse mutations for logging in and logging out
- Adds missing `username` attribute to `UserFactory` and GraphQL Schema
- Adds authentication functionality to the Quasar web form at `/login` with Apollo Client and Apollo Links
- Changes authentication parameters from username/password combo to email/password combo

**Acceptance Testing**

1. Spin up CCR locally
2. In a command line, run `lando artisan tinker` to enter the Psy Shell of the database container and `App\Models\User::factory()->count(1)->create();` to generate a user
3. Note the email of the generated user
4. Visit the CCR local URL within a browser and click the `LOGIN` link to navigate to `/login`
5. Verify in the browser's developer tools Storage tab (or equivalent) that no cookies are stored within the browser 
6. In the web form, enter the email address of the generated user with a password that is incorrect (ex: `wrongpassword`)
7. Verify in the browser's developer tools Storage tab (or equivalent) that no cookies are within the browser still and that an internal server error was logged to the browser Console
8. In the web form, enter the email address of the generated user with the default password: `password`
9. Verify in the browser's developer tools Storage tab (or equivalent) that a CSRF mitigation cookie and a session cookie are stored within the browser upon a successful authentication and that the user's details were logged to the browser Console

PR #101, opened 2021-01-20
fix: merge migrations in graphql schema

wreality:
#97 introduced a regression into the graphql schema file.  (multiple root types aren't allowed).  This PR merges the two into a single root type.

### DIFF
--- a/backend/graphql/schema.graphql
+++ b/backend/graphql/schema.graphql
@@ -60,9 +60,6 @@ type User {
     updated_at: DateTime
 }
 
-"""
-Create a new user.
-"""
 type Mutation {
     "Create a new user"
     createUser(user: CreateUserInput! @spread): User! @create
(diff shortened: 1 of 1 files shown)


backend/graphql/schema.graphql
```diff
@@ -64,13 +64,12 @@ type User {
 Create a new user.
 """
```
gmeben, line 65: Shouldn't we remove this comment if it's being added within the `Mutation` object?

```suggestion

```

Side Note: I tried doing [a multiline suggestion](https://github.blog/changelog/2020-02-26-multi-line-code-suggestions-beta/) but it's not working for me.

gmeben (changes requested):

backend/graphql/schema.graphql
wreality: You are correct. 👍

wreality:
Thanks for the catch. I'm good with it, go ahead and approve and merge if you're happy as well.

gmeben (approved):

Issue #103, opened 2021-01-21
An SQL error occurs if the name field is not supplied for the createUser migration


wreality:
**Problem/Why**

An SQL error occurs if name field is not supplied to the createUser migration.

---

**Goal/What**

As name is a nullable field, createMigration should succeed if name is not supplied and all other fields pass validation. 

---

**Solution/How**

The error references a lack of a default value for name in the database. A DB migration should be created that either allows the field to be nullable or defaults to an empty string. It would also be appropriate to implement unit testing to describe and test for the resolution of the bug. 

---
### To Do / Acceptance Criteria
- [x] Create unit tests
- [x] Add migration to make field nullable
### Review Checklist
- [ ] Review Unit test
- [ ] Confirm tests pass `lando artisan test` (lando rebuild may be required)

PR #109, opened 2021-01-22
fix(api): fix error when name field omitted in createUser

wreality:
ref #103


backend/database/migrations/2021_01_22_061707_allow_user_name_null.php
```diff
@@ -0,0 +1,32 @@
...
+
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
+    {
+        Schema::table('users', function (Blueprint $table) {
+            $table->string('name')->nullable(false)->change();
```
gmeben, line 29: How does this work? 

If the `down()` methods of all the migrations are ran chronologically, the `users` table won't exist to be modified by the time this `down()` method is ran. Doesn't that make this redundant?

Or is it possible that this `down()` method can be ran independent of the others? (new to me)

backend/tests/Feature/CreateUserEndpointTest.php
```diff
@@ -0,0 +1,49 @@
...
+use Nuwave\Lighthouse\Testing\MakesGraphQLRequests;
+use Illuminate\Foundation\Testing\RefreshDatabase;
+use Tests\TestCase;
+
+class CreateUserEndpointTest extends TestCase
+{
+    use MakesGraphQLRequests;
+    use RefreshDatabase;
+    /**
+     * A basic feature test example.
```
gmeben, line 14: This comment is too generic and the function name is already descriptive enough of what it does.
```suggestion

```

backend/tests/Feature/CreateUserEndpointTest.php
```diff
@@ -0,0 +1,49 @@
...
+                    username
+                }
+            }
+        ');
+
+        $response->assertJsonPath("data.createUser.username", "testusername");
+
+
+    }
+
```
gmeben, line 33: Add the missing DocBlock here:
```suggestion
    /**
     * @return void
     */
```

backend/tests/Feature/CreateUserEndpointTest.php
```diff
@@ -0,0 +1,49 @@
...
+namespace Tests\Feature;
+
+use Nuwave\Lighthouse\Testing\MakesGraphQLRequests;
+use Illuminate\Foundation\Testing\RefreshDatabase;
+use Tests\TestCase;
+
+class CreateUserEndpointTest extends TestCase
+{
+    use MakesGraphQLRequests;
+    use RefreshDatabase;
```
gmeben, line 12: I like having a little separation between functions/classes and use statements. This convention seems to be demonstrated everywhere else in the code too.
```suggestion
    use RefreshDatabase;

```

backend/tests/Feature/CreateUserEndpointTest.php
```diff
@@ -0,0 +1,49 @@
...
+        
+            mutation {
+                createUser(user: {email: "brandnew@gmail.com", password: "KajSu8viptUrz&", username: "testusername"}) {
+                    username
+                }
+            }
+        ');
+
+        $response->assertJsonPath("data.createUser.username", "testusername");
+
```
gmeben, line 46: Might be useful to also assert that `data.createUser.name` is null. I'll see if I can come up with something.

backend/tests/Feature/CreateUserEndpointTest.php
```diff
@@ -0,0 +1,49 @@
...
+        
+            mutation {
+                createUser(user: {name: "", email: "brandnew@gmail.com", password: "KajSu8viptUrz&", username: "testusername"}) {
+                    username
+                }
+            }
+        ');
+
+        $response->assertJsonPath("data.createUser.username", "testusername");
+
```
gmeben, line 30: Might be useful to also assert that `data.createUser.name` is empty. I'll see if I can come up with something.

gmeben (changes requested):

backend/database/migrations/2021_01_22_061707_allow_user_name_null.php
wreality: The down methods would be run in reverse chronological order. It's the undo of up. Also you can just undo a single migration (artisan migrate:rollback --step=1)

backend/tests/Feature/CreateUserEndpointTest.php
wreality: Yep that's the default, thanks for catching it!

backend/tests/Feature/CreateUserEndpointTest.php
gmeben: I added an assertion to check that the value of this property is an empty string. Let me know what you think.

backend/tests/Feature/CreateUserEndpointTest.php
gmeben: I added an assertion to check that the value of this property is null. Let me know what you think.

backend/tests/Feature/CreateUserEndpointTest.php
wreality: Added in a later commit.

backend/tests/Feature/CreateUserEndpointTest.php
wreality: Added.

backend/tests/Feature/CreateUserEndpointTest.php
wreality: Added.

backend/tests/Feature/CreateUserEndpointTest.php
wreality: Resolved

gmeben (approved):

PR #114, opened 2021-01-26
build: Install Laravel permissions

gmeben:
This PR

* Installs the third-party dependency `spatie/laravel-permissions`
* Adds `Role` and `Permission` models
* Adds some unit tests to ensure it works as expected

**Acceptance Testing**

* In the command line, navigate to the root directory of the locally cloned CCR project
* Checkout this branch `laravel-permission` in git
* Run `lando rebuild` to ensure the new dependency is installed
* Navigate to the `backend` directory
* Run tests with `lando artisan test --testsuite=Feature --filter=UserPermissionsTest`

### DIFF
--- a/backend/tests/Feature/UserPermissionsTest.php
+++ b/backend/tests/Feature/UserPermissionsTest.php
@@ -106,4 +106,25 @@ class UserPermissionsTest extends TestCase
         $user->assignRole('Application Administrator');
         $this->assertTrue($user->can($name));
     }
+
+    /**
+     * @return void
+     */
+    public function testUserDoesNotHavePermissionToCreatePublicationsByAssignedRole()
+    {
+        $name = 'create publications';
+        $user = User::factory()->create();
+        $permission = Permission::factory()->create([
+            'name' => $name
+        ]);
+        $role_1 = Role::factory()->create([
+            'name' => 'Application Administrator'
+        ]);
+        $role_2 = Role::factory()->create([
+            'name' => 'Publication Administrator'
+        ]);
+        $role_1->givePermissionTo($name);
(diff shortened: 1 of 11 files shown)


backend/tests/Feature/UserPermissionsTest.php
```diff
@@ -0,0 +1,58 @@
...
+        $name = 'Application Administrator';
+        $user = User::factory()->create();
+        $role = Role::factory()->create([
+            'name' => $name
+        ]);
+        $user->assignRole($name);
+        $this->assertNotNull($role->name);
+        $this->assertNotEmpty($role->name);
+        $this->assertIsString($role->name);
+        $this->assertEquals($role->name, $name);
```
wreality, line 55:
```suggestion
```
You've already tested that creating a role works above.  I'd suggest keeping the assertions limited to what you're trying to test as a habit.  Help avoid brittle tests.

backend/tests/Feature/UserPermissionsTest.php
```diff
@@ -0,0 +1,58 @@
...
+     */
+    public function testCreationOfApplicationAdministratorRole()
+    {
+        $name = 'Application Administrator';
+        $role = Role::factory()->create([
+            'name' => $name
+        ]);
+        $this->assertNotNull($role->name);
+        $this->assertNotEmpty($role->name);
+        $this->assertIsString($role->name);
```
wreality, line 37:
```suggestion
```
These are all implied by the assertEquals below.. which is what you really want to test.

wreality (approved):
Looks good to me!  And thanks for adding unit tests!

PR #115, opened 2021-01-26
feat: upgrade lighthouse to version 5.1

wreality:
Upgrades lighthouse to version 5.1.  Only one minor change has been identified in our code base and `artisan test` completes successfully.

### DIFF
--- a/backend/config/lighthouse.php
+++ b/backend/config/lighthouse.php
@@ -210,7 +210,7 @@ return [
     |
     */
 
-    'debug' => \GraphQL\Error\Debug::INCLUDE_DEBUG_MESSAGE | \GraphQL\Error\Debug::INCLUDE_TRACE,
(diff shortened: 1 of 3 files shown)


backend/config/lighthouse.php
```diff
@@ -210,7 +210,7 @@
     |
     */
 
-    'debug' => \GraphQL\Error\Debug::INCLUDE_DEBUG_MESSAGE | \GraphQL\Error\Debug::INCLUDE_TRACE,
+    'debug' => env('LIGHTHOUSE_DEBUG', \GraphQL\Error\DebugFlag::INCLUDE_DEBUG_MESSAGE | \GraphQL\Error\DebugFlag::INCLUDE_TRACE),
```
gmeben, line 213: This makes me wonder if we should have documentation somewhere for `.env` variables, unless it already exists?

gmeben (approved):

PR #118, opened 2021-01-28
feat: Add user roles

gmeben:
This PR 

* Populates user roles to the `roles` table so they can be used formally throughout CCR
* Extends the `Role` model
    * Adds constants for the names of the user roles
    * Adds a static method for getting an array list of all user roles
* Fixes testing around user roles and permissions

**Unit Testing**

Run the new unit tests to verify things work correctly

**Acceptance Testing**

1. In the `backend` directory of this repo, run `lando artisan migrate:fresh`
2. Run `lando mysql laravel`
3. Run `select * from roles` and verify that all roles have a corresponding row with columns that have proper data populated within them (i.e. "web" as the guard name and timestamps for the date-related columns)

### DIFF
--- a/backend/database/migrations/2021_01_27_213544_insert_user_roles.php
+++ b/backend/database/migrations/2021_01_27_213544_insert_user_roles.php
@@ -3,7 +3,6 @@
 use Illuminate\Database\Migrations\Migration;
 use Illuminate\Support\Facades\Schema;
 use Illuminate\Support\Facades\DB;
-use App\Models\Role;
 
 class InsertUserRoles extends Migration
 {
@@ -14,8 +13,15 @@ class InsertUserRoles extends Migration
      */
     public function up()
     {
-        $roles = Role::getArrayOfAllRoleNames();
         $guard_name = config('auth.defaults.guard');
+        $roles = [
+            'Application Administrator',
+            'Publication Administrator',
+            'Editor',
+            'Review Coordinator',
+            'Reviewer',
+            'Submitter'
+        ];
 
         foreach ($roles as $role) {
             DB::table('roles')->insert([
--- a/backend/database/migrations/2021_01_27_213544_insert_user_roles.php
+++ b/backend/database/migrations/2021_01_27_213544_insert_user_roles.php
@@ -2,8 +2,6 @@
 
 use Illuminate\Support\Facades\DB;
 use Illuminate\Database\Migrations\Migration;
-use Illuminate\Database\Schema\Blueprint;
-use Illuminate\Support\Facades\Schema;
 use App\Models\Role;
 
 class InsertUserRoles extends Migration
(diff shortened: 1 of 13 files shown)


wreality (changes requested):
I wonder if adding records to the DB would be better in seeders rather than migrations.  What do you think?: https://laraveldaily.com/10-tips-about-data-seeding-in-laravel/

My opinion is that migrations should always be individually reversible and should be runnable even as the code base changes over time.  Seeders on the other hand are all about setting up the default data.

I'd happily talk through my thoughts if you like.  I'm not steadfast in this opinion and would totally consider changing it.

gmeben:
@wreality I did consider using seeding, but I don't believe seeding is for setting up default data that's meant to persist to production. According to Laravel's documentation, seeding is meant for generating test data for testing purposes: https://laravel.com/docs/8.x/seeding

gmeben:
@wreality You make a good point about the migration needing to be individually reversible, so I'll extend the migration to delete those records on the `down()` method.

backend/database/migrations/2021_01_27_213544_insert_user_roles.php
```diff
@@ -0,0 +1,36 @@
...
+     * @return void
+     */
+    public function up()
+    {
+        $roles = Role::getArrayOfAllRoleNames();
+        foreach ($roles as $role) {
+            Role::create([
+                'name' => $role
+            ]);
+        }
```
wreality, line 22:
```suggestion
        foreach ($roles as $index => $role) {
            DB::table('roles')->insert(
                [
                    'id' => $index + 1,
                    'name' => $role,
                    'guard_name' => 'web',
                    'created_at' => date('Y-m-d H:i:s'),
                    'updated_at' => date('Y-m-d H:i:s')
                ]
            );
        }
```

backend/database/migrations/2021_01_27_213544_insert_user_roles.php
```diff
@@ -0,0 +1,36 @@
...
+class InsertUserRoles extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        $roles = Role::getArrayOfAllRoleNames();
```
wreality, line 17: I know it's less DRY, but the roles to be created should be explicit here rather than pulled from the Roles class.

wreality (changes requested):
Agree with doing the inserts in the migration rather than seeds.  However we shouldn't reference the Role class in the migration.

backend/database/migrations/2021_01_27_213544_insert_user_roles.php
gmeben: Done. I removed the dependence on the `Role` class here.

backend/database/migrations/2021_01_27_213544_insert_user_roles.php
gmeben: Refactored to an insert

wreality (approved):

wreality:
@gmeben Did you want me to merge this or were you going to ?

gmeben:
@wreality I was going to ask you the same thing. I'll merge.

PR #123, opened 2021-02-02
feat: extend graphql with user roles

gmeben:
This PR extends the GraphQL endpoint with user roles.

### DIFF
--- a/backend/graphql/schema.graphql
+++ b/backend/graphql/schema.graphql
@@ -95,7 +95,7 @@ type User {
     username: String!
     created_at: DateTime!
     updated_at: DateTime
-    roles: [Role]
+    roles: [Role!]!
 }
 
 type Mutation {
--- a/backend/graphql/schema.graphql
+++ b/backend/graphql/schema.graphql
@@ -15,6 +15,7 @@ type Query {
     "Validate details of a new user.  Returns true if supplied user fields would be valid."
     validateNewUser(user: ValidateNewUserInput): Boolean @returnTrue
 
+    "Return pre-defined user roles in the application"
     role(id: ID @eq): Role @find
 }
 
--- a/backend/graphql/schema.graphql
+++ b/backend/graphql/schema.graphql
@@ -73,6 +73,9 @@ input CreateUserInput {
         @hash
 }
 
+"""
+A user role for permissions
+"""
 type Role {
     id: ID!
     name: String!
--- a/backend/graphql/schema.graphql
+++ b/backend/graphql/schema.graphql
@@ -14,6 +14,8 @@ type Query {
 
     "Validate details of a new user.  Returns true if supplied user fields would be valid."
     validateNewUser(user: ValidateNewUserInput): Boolean @returnTrue
+
+    role(id: ID @eq): Role @find
 }
 
 "Validate the availability of username and email via the validateNewUser query"
@@ -71,6 +73,14 @@ input CreateUserInput {
         @hash
 }
 
+type Role {
+    id: ID!
+    name: String!
+    guard_name: String!
+    created_at: DateTime
+    updated_at: DateTime
+}
+
 """
 A user account.
 """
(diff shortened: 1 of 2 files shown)


backend/graphql/schema.graphql
```diff
@@ -81,6 +91,7 @@ type User {
     username: String!
     created_at: DateTime!
     updated_at: DateTime
+    roles: [Role]
```
wreality, line 94:
```suggestion
    roles: [Role!]!
```

backend/tests/Feature/UserPermissionsTest.php
```diff
@@ -126,4 +127,82 @@ public function testUserDoesNotHaveTestPermissionByAssignedRole()
...
+                'name' => 'Test User Role'
+            ]
+        ];
+        $getUserResponse->assertJsonPath("data.user.roles", $expected_array);
+    }
+
+    /**
+     * @return void
+     */
+    public function testRoleIsQueryableFromUserCreatedViaFactory()
```
wreality, line 180: And now I see that you have one created by the factory as well.  I guess I'm curious what we're testing differently btw these two tests.

wreality (changes requested):

backend/tests/Feature/UserPermissionsTest.php
```diff
@@ -126,4 +127,82 @@ public function testUserDoesNotHaveTestPermissionByAssignedRole()
...
+                    password: "KajSu8viptUrz&",
+                    username: "testusername",
+                    name: "Test Name"
+                }) {
+                    id
+                    name
+                    username
+                }
+            }'
+        );
```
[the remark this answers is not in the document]
gmeben: I think it's useful to test that the same information is available when creating users in different techniques.

backend/graphql/schema.graphql
```diff
@@ -71,6 +73,14 @@ input CreateUserInput {
         @hash
 }
```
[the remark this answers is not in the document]
gmeben: Added

backend/graphql/schema.graphql
```diff
@@ -14,6 +14,8 @@ type Query {
 
     "Validate details of a new user.  Returns true if supplied user fields would be valid."
     validateNewUser(user: ValidateNewUserInput): Boolean @returnTrue
+
```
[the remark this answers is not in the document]
gmeben: Added

backend/graphql/schema.graphql
gmeben: Roles are optional at the moment. I suspect we'll have to do more work elsewhere to make them required before I can justify designating the return of null value here as an invalid request.

backend/tests/Feature/UserPermissionsTest.php
gmeben: It's a sanity check. If a user was created via User factory or via Lighthouse, we want to ensure that the data exists and functions the same regardless.

backend/graphql/schema.graphql
gmeben: I just tried querying a user that has no role with `[Role!]!` set in the schema and observed no different to `[Role]` so now I'm not even sure what designating this as a non-nullable field is actually doing.

wreality: We don't currently return null, we return an empty array. 

https://stackoverflow.com/questions/46770501/graphql-non-nullable-array-list/46771890

backend/tests/Feature/UserPermissionsTest.php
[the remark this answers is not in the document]
gmeben: This test was removed

backend/tests/Feature/UserPermissionsTest.php
gmeben: The duplicate test was removed.

wreality (approved):

PR #128, opened 2021-02-10
feat: add first user permission

gmeben:
This PR adds the first permission formally to the system: `update user for others`

Review Checklist

- [x] The new permission exists in the permissions table
- [x] PHPUnit tests successfully run locally


wreality:
I feel like we're missing how the association between permission and roles that (at least by our design) will have that permission.

This is the weirdness of the roles and permissions being records in the DB for me.  I think it's right, but I'm used to these things being defined in the code base.  I'm going to ponder how to do this.

gmeben:
Yeah, this is a weird position for sure. There's pros and cons to either strategy. I think there's tremendous benefit to using the package's API in migrations (like the article you linked demonstrates), but I also like knowing that the migrations can be run independent of any third-party code. Although, I can't speak to the exact benefit of being able to run them independently in the long term. I suppose I lean toward using the package's API which is likely why I originally coded it that way.

backend/database/migrations/2021_01_27_213544_insert_user_roles.php
```diff
@@ -40,8 +27,11 @@ public function up()
...
     {
-        Schema::disableForeignKeyConstraints();
-        DB::table('roles')->truncate();
-        Schema::enableForeignKeyConstraints();
+        $roles = Role::getArrayOfAllRoleNames();
+        foreach ($roles as $role) {
+            /** @var $role Role */
+            $role = Role::findByName($role);
+            $role->delete();
+        }
```
wreality, line 35: Thanks for this.  I like this a lot better than relying on truncate to do the job.

wreality (approved):

PR #135, opened 2021-02-17
feat: Issue 120 update user backend

### LINKED ISSUE #120: Add updateUser GraqphQL mutation.
**Problem/Why**
---

The GraphQL endpoint does not have the ability to update user fields.

**Goal/What**
---

An API user should be able to update their username, email address, name, and password while logged into their account.

**Solution/How**
---

Create GraphQL endpoints for updating the user's information.

-   [x] New graphql endpoint for updating user

### Review Checklist
-   An API user can update their:
  - [ ] Username
  - [ ] Password
  - [ ] Email address
  - [ ] Name
-   [ ] Appropriate unit tests

gmeben:
closes #120 

This prepares the backend for user updates via mutations on the GraphQL API

**Unit Testing**

1. On the cloned repo, start Lando `lando start`
2. Navigate to the backend directory `cd backend`
3. Run the unit tests `lando artisan test`


backend/app/Policies/UserPolicy.php
```diff
@@ -0,0 +1,105 @@
...
+        if (!$is_updating_own_user
+            && $user->can(Permission::UPDATE_USERS)
+        ) {
+            return true;
+        }
+
+        // TODO: Check if user can update user within own publication
+
+        return $is_updating_own_user;
+
```
wreality, line 67:
```suggestion
        //User is updating their own record.
        if ($user->id === $model->id) {
            return true;
        }
        //User has global permission to update users
        if ($user->can(Permission::UPDATE_USERS)) {
            return true;
        }

        // TODO: Check if user can update user within own publication
        // No explicit permission so return false.
        return false;
```
I feel like this is a little easier to digest.

backend/app/Providers/AuthServiceProvider.php
```diff
@@ -15,6 +17,7 @@ class AuthServiceProvider extends ServiceProvider
      */
     protected $policies = [
         // 'App\Models\Model' => 'App\Policies\ModelPolicy',
+        User::class => UserPolicy::class
```
wreality, line 20: I didn't realize this was a step for applying policies.  I like it. (It had been feeling a little too "magic" for my taste)

wreality (changes requested):
@gmeben I only have one refactor to suggest.

backend/app/Providers/AuthServiceProvider.php
gmeben: Your feeling wasn't wrong! https://laravel.com/docs/8.x/authorization#policy-auto-discovery

wreality (approved):

Issue #139, opened 2021-02-19
Fix console error on page refresh


wreality:
**Problem/Why**

There is a console error reported on page load: 
![image.png](https://images.zenhubusercontent.com/5fe17935da86cae135bcd2ab/37dcb9b6-9dd5-411e-bbe8-5e1b939d9b45)

---

**Goal/What**

There shouldn't be an error...

---

**Solution/How**

Remove the calls to the vuex store from App.vue (vuex store is not used and has been removed, this is a regression).  Also check that we're not trying to use vuex anywhere else.

---

### To Do / Acceptance Criteria

- [x] Fix error in console.
- [ ] Check/remove other vuex calls

### Review Checklist

- [x] No more error in console

PR #153, opened 2021-02-26
fix: resolve console error on page load

wreality:
Removes call to vuex store (we're no longer using vuex for state management).  Also checked the rest of the codebase and wasn't able to find any other calls to vuex remaining.

See connected issue #139  for review/acceptance criteria


gmeben (approved):
Seems fine. All tests run successfully on my machine. I was able to reproduce the error on the master branch while loading the landing page. I see that it's resolved on this branch.

Issue #124, opened 2021-02-05
Add cypress tests to register and login views


wreality:
**Problem/Why**

The `/register` and `/login` views are currently not tested in cypress, leading to the possibility that regressions may not be caught in the future.

---

**Goal/What**

The `/register` and `/login` views should have tests in the cypress test suite.

---

**Solution/How**

Implement integration tests in cypress.

---

#### In `client/test/cypress/integration/register.spec.js`
-   [ ] Write test to confirm that a user can be created by typing into the form
-   [ ] Confirm that the form redirects to `/dashboard` on success and that the user is logged in
-   [ ] Write test to confirm that creation fails when username or email already exists
#### In `client/test/cypress/support/commands.js`
Write commands for future tests to use to login, logout, and create a user using ApolloClient (use cy.getVue() to access the root Vue component).
-  [ ] Write a command to create a user.  The command should ignore user already exists errors. 
-  [ ] Write a command to login a user.
-  [ ] Write a command to logout a user.

#### In `client/test/cypress/integration/login.spec.js`
-  [ ] Use the create user helper to create a test user.
-  [ ] Write a test to confirm login success with the correct user name and password.
-  [ ] Confirm redirect to `/dashboard` on login success
-  [ ] Confirm `/logout` logs the user out.
-  [ ] Confirm that visiting a protected route `/account/profile` redirects to `/login` with a banner message.
-  [ ] Confirm that following login the user is redirected to their original destination `/account/profile`
-  [ ] Log the test user out (using logout helper)
-  [ ] Confirm that logging in with an incorrect password fails with an error message.

### Review Checklist

-   [ ] Tests implemented and passing

PR #157, opened 2021-03-02
test: add integration tests for login and register

### LINKED ISSUE #151: Enter key does not trigger login from password field
**Problem/Why**

In the login form, when in the password field, the enter key does not trigger form submit.  This annoys me.

---

**Goal/What**

When the password field has focus, pressing enter should trigger form submission.

---

**Solution/How**

Implement the enter key event listener (or another feature of Vue that allows the form submission event to be triggered via the enter key).  It may be that the event listener should be on the submit event of the form rather than the click event of the button.  This may also be better for accessibility, but I'm not sure if that's the case.

---

- [ ] Make enter key work in the password field on the login page.

### Review Checklist

- [ ] Login key works in the password field on the login page.

wreality:
See #124 for acceptance/review criteria

Also closes #151


client/test/cypress/integration/login.spec.js
```diff
@@ -0,0 +1,54 @@
...
+    });
+
+    it('redirects to login when requesting a protected page', () => {
+        cy.visit('/account/profile');
+        cy.url().should('include', '/login');
+        cy.get('[role="alert"]').contains('login to access that page');
+
+        cy.get('.q-form').within(() => {
+            cy.dataCy('email_field').type('regularuser@ccrproject.dev');
+            cy.dataCy('password_field').type('regularPassword!@#{enter}');
```
gmeben, line 50: `{enter}` is nice

gmeben (approved):

Issue #161, opened 2021-03-04
fix: Lint/Style Checking on 0.9.0 CHANGELOG


gmeben:
### Problem/Why
A markdown linting failure occurred with the release of the CHANGELOG for release 0.9.0: https://github.com/MESH-Research/CCR/runs/2027621976

### Goal/What
CHANGELOG.md should not have linting errors and automated tools should generate files that pass linting.

### Solution/How
Fix the listed errors in a new branch and update build tools to generate well-formatted files.

### To Do / Acceptance Criteria

- [x] Fix errors
- [x] Update build tools.
- [x] Update docs re: how to create a release

### Review Checklist

- [x] Linting errors should no longer be present

wreality:
I noticed this last night, but hadn't figured out how to deal with it yet. The change log is autogenerated and I don't want to have to manually fix it every time. 

It's also the same changelog format that a wide number of packages use now so I'm curious as to why it generates a lint erroring changelog. 

I'm investigating options.

wreality:
N.b. This doesn't require a release since the changes don't impact the application.

gmeben:
@wreality Do we not consider the application's documentation as part of the application? It lives in the same repository.

PR #162, opened 2021-03-08
build: correct changelog update build process

wreality:
Updates the build process to address #161 and fixes current errors in the changelog.

Also, updates documentation to add instructions for tagging a release.


scripts/ccr-custom-changelog/release.js
```diff
@@ -0,0 +1,43 @@
...
+const config = {
+	types: [
+		{ type: "feat", section: "Features" },
+		{ type: "fix", section: "Bug Fixes" },
+		{ type: "perf", section: "Performance Improvements", hidden: false },
+		{ type: "test", section: "Test Suite", hidden: false },
+		{ type: "build", section: "Build System", hidden: false },
+		{ type: "docs", section: "Documentation", hidden: false },
+		{ type: "style", section: "Other", hidden: false },
+		{ type: "refactor", section: "Other", hidden: false },
```
gmeben, line 14: I've been using `ref` as a prefix in commits for refactors. Is this suggested addition appropriate?
```suggestion
		{ type: "ref", section: "Other", hidden: false },
		{ type: "refactor", section: "Other", hidden: false },
```

gmeben (approved):

scripts/ccr-custom-changelog/release.js
wreality: We could change it but that would also mean adding it to the allowed types in the linter. Also I think ref is more ambiguous than refactor. (Reference, refactor, reformat, refine)

Issue #147, opened 2021-02-19
Add a SystemAdmin route group to client routes


wreality:
**Problem/Why**

As new system admin features are added they will each need to enforce system admin roles to access them.

---

**Goal/What**

To avoid duplication of these checks, a route group should be created in `routes.js` to check that the user has the appropriate role/permission.

---

**Solution/How**

### To Do
- [x] Update `client/src/apollo/apollo-client-hooks.js` to implement a router meta guard for roles and or permissions
- [x] Update `client/src/router/routes.js` to add a route group for systemadmin pages.
- [x] Create placeholder page inside systemadmin child routes to test implementation
- [x] CYPRESS: Test that a user without admin role is shown an error page.
- [x] CYPRESS: Test that a user with admin role is able to visit page

### Acceptance Criteria

- [ ] Tests are complete and passing
- [ ] Placeholder route is only accessible to a system admin user.

### Review Checklist

- [ ] Tests are complete and passing.

PR #164, opened 2021-03-09
refactor(client): add roles based router guard

wreality:
Addresses #147 by adding a new router guard.


backend/database/seeders/DatabaseSeeder.php
```diff
@@ -19,8 +21,15 @@ public function run()
...
             'name' => 'Regular User',
-            'password' => Hash::make('regularPassword!@#')
+            'password' => Hash::make('regularPassword!@#'),
         ]);
-        
+
+        $user = User::factory()->create([
+            'username' => 'systemAdminUser',
+            'email' => 'adminuser@ccrproject.dev',
+            'name' => 'System Admin User',
```
gmeben, line 30: Should we change the term from 'Application Admin' to 'System Admin'? 

I notice other team members keep defaulting to "system" when describing this role as well.

client/src/apollo/apollo-client-hooks.jest.spec.js
```diff
@@ -0,0 +1,195 @@
...
+
+        await beforeEachRequiresAuth(apolloMock, to, undefined, next);
+
+        expect(apolloMock.query).toHaveBeenCalled();
+        expect(setSession).toHaveBeenCalled();
+        expect(next).toHaveBeenCalled();
+        expect(next.mock.calls[0][0]).not.toBe(undefined);
+    });
+
+    it('does nothing it route does not require auth', async () => {
```
gmeben, line 64: Need to re-write this description.

client/src/apollo/apollo-client-hooks.jest.spec.js
```diff
@@ -0,0 +1,195 @@
...
+        const next = jest.fn();
+
+        await beforeEachRequiresRoles(apolloMock, to, undefined, next);
+
+        expect(apolloMock.query).toHaveBeenCalled();
+        expect(next).toHaveBeenCalled();
+        expect(next.mock.calls[0][0]).not.toBe(undefined);
+    });
+
+    it('does nothing is requireRoles is not present', async () => {
```
gmeben, line 134: Re-write this description please.

client/src/apollo/apollo-client-hooks.jest.spec.js
```diff
@@ -0,0 +1,195 @@
...
+        await beforeEachRequiresRoles(apolloMock, to, undefined, next);
+
+        expect(apolloMock.query).not.toHaveBeenCalled();
+        expect(next).toHaveBeenCalled();
+        expect(next.mock.calls[0][0]).toBe(undefined);
+    });
+
+    it('requires all nested roles', async () => {
+        const to = {
+            matched: [{ meta: { requiresRoles: 'admin' } }, {meta: { requiresRoles: 'specialAdmin' } }]
```
gmeben, line 150: What does a "specialAdmin" represent?

gmeben (changes requested):

client/src/apollo/apollo-client-hooks.jest.spec.js
wreality: a random role that's not really a thing.

backend/database/seeders/DatabaseSeeder.php
wreality: Whether it's application or system doesn't really matter to me.  I can go either way.  I will say that I think the role titles should be APPLICATION_ADMIN rather than "Application Admin" (caps snakecase over human strings)

wreality: I updated the test user to be title application administrator.

client/src/apollo/apollo-client-hooks.jest.spec.js
gmeben: What do you think about renaming this `specialTestAdmin` just to be clear about the purpose of this object?

wreality: I changed the name in the latest commit.

Issue #8, opened 2020-06-07
Add database schema and GraphQL endpoints for user profile data


wreality:
**Problem/Why**

There is currently not a way for users to describe metadata about themselves into the applicaiton.

---

**Goal/What**

The user database should include profile data about users.  Reference the metadata collected in #64 for the types of information that should be collected.

---

**Solution/How**

The implementation in the database should be flexible so as to not require database migrations to add user metadata fields in the future, especially as these fields do not need to be queried / joined individually.  If there _are_ some fields that would be joined / filtered against, those should be added as columns directly.

---

### To Do / Acceptance Criteria

- [x] Create a new JSON column on the users table for storing user profile metadata
- [x] Add the user profile metadata field to the user type in the graphql schema
- [x] Add the user profile metadata field to the existing updateUser mutation
- [x] Implement tests to ensure the data can be queried and mutated.
- [x] Implement profile metadata validation (only accept known fields, and set reasonable max length limits)

### Review Checklist

- [x] Tests pass
- [x] User profile data is stored and associated with the user.
- [x] User profile fields are flexible.

gmeben:
There's no built in JSON scalar type in GraphQL as far as I can tell. When defining the metadata column in the GraphQL schema as a `String` type, queries represent the stored JSON with escaped quotes in the result regardless of the stored JSON containing escaped quotes: 

```JSON
{
  "data": {
    "users": {
      "data": [
        {
          "id": "1",
          "profile_metadata": "{\"Phone\": \"444-111-9999\", \"Salutation\": \"mrs.\", \"Mailing Address\": {\"City\": \"Seattle\", \"State\": \"WA\", \"Postal Code\": \"98052\", \"Street Address\": \"20341 Whitworth Institute 405 N. Whitworth\"}}"
        }
      ]
    }
  }
}
```

If we want results to represent this field as a fully-parsed JSON object, like this...

```
{
  "data": {
    "users": {
      "data": [
        {
          "id": "1",
          "profile_metadata": {
            "Salutation": "mrs.",
            "Phone": "444-111-9999",
            "Mailing Address": {
              "City": "Seattle",
              "State": "WA",
              "Postal Code": "98052",
              "Street Address": "20341 Whitworth Institute 405 N. Whitworth"
            }
          }
        }
      ]
    }
  }
}
```

...then we'll likely need to implement this third party package: [graphql-type-json](https://github.com/taion/graphql-type-json)

[...]

PR #168, opened 2021-03-10
feat(api): add profile metadata to user type

gmeben:
This prepares the backend for validating and receiving query and mutation requests on user profile metadata.

Bonus: adds a link to the Account Profile page to the sidebar menu on the frontend.

Closes Issue #8


backend/graphql/schema.graphql
```diff
@@ -33,6 +33,50 @@ input ValidateNewUserInput @validator(class: "UserInputValidator") {
     email: String
 }
 
+"""
+Input type for adding social media data to profile metadata
+"""
+input UpdateProfileMetadataSocialMediaInput @validator(class: "ProfileMetadataSocialMediaValidator") {
```
wreality, line 39: Can we use namespaces in the class property?  I.e. would it be better to be able to make `App\Graphql\Validators\ProfileMetadata\SocialMediaValidator` to keep the validators organized?  I don't know if the class property will happily accept a namespaced class or not though.

backend/app/GraphQL/Validators/ProfileMetadataSocialMediaValidator.php
```diff
@@ -0,0 +1,44 @@
+<?php
+declare(strict_types=1);
+
+namespace App\GraphQL\Validators;
+
+use Nuwave\Lighthouse\Validation\Validator;
+
+class ProfileMetadataSocialMediaValidator extends Validator
```
wreality, line 8: Related to the similar comment on schema.graphql.  Would these be better organized inside a namespace (if the @validator directive will allow us to specify a namespace).  i.e. `class SocialMediaValidator` inside `Validators/ProfileMetadata`.

backend/graphql/schema.graphql
```diff
@@ -68,6 +112,9 @@ input UpdateUserInput @validator(class: "UserInputValidator") {
 
     "Password used to login to the application."
     password: String @hash
+
+    "Arbitrary metadata related to the profile of a user."
```
wreality, line 116:
```suggestion
    "Metadata related to the profile of a user."
```

backend/graphql/schema.graphql
```diff
@@ -106,6 +153,42 @@ type User {
     email_verified_at: DateTime
     roles: [Role!]! @belongsToMany
     permissions: [Permission!] @belongsToMany
+    profile_metadata: ProfileMetadata
+}
+
```
wreality, line 158: Individual subtypes should have a docstring on them.  Every field would be overkill, but at least the subtypes.

backend/tests/Feature/UserQueryTest.php
```diff
@@ -0,0 +1,122 @@
...
+
+class UserQueryTest extends TestCase
+{
+    use MakesGraphQLRequests;
+    use RefreshDatabase;
+
+    /**
+     * @return array
+     */
+    public function profileMetadataProvider()
```
wreality, line 19:
```suggestion
    public function profileMetadataProvider(): array
```

backend/tests/Feature/UserQueryTest.php
```diff
@@ -0,0 +1,122 @@
...
+                ],
+            ],
+        ];
+    }
+
+    /**
+     * @dataProvider profileMetadataProvider
+     * @return void
+     */
+    public function testThatUserDetailsCanBeQueried(?array $profile_metadata): void
```
wreality, line 69:
```suggestion
    public function testThatUserDetailsCanBeQueried(array $profile_metadata): void
```

backend/tests/Feature/UserQueryTest.php
```diff
@@ -0,0 +1,122 @@
...
+            }',
+            [ 'id' => $user->id ]
+        );
+        $response->assertJson([
+            'data' => [
+                'user' => [
+                    'profile_metadata' => $profile_metadata,
+                ],
+            ],
+        ]);
```
wreality, line 120: I tweaked this to be a little less repetitive in this case.  Feel free to revert if you have a strong opinion that I should remain the way it was.

client/src/layouts/MainLayout.vue
```diff
@@ -23,6 +23,17 @@
...
+            v-ripple
+            to="/account/profile"
+          >
+            <q-item-section avatar>
+              <q-icon name="account_circle" />
+            </q-item-section>
+            <q-item-section>
+              {{ $t("header.account_link") }}
+            </q-item-section>
+          </q-item>
```
wreality, line 36: I don't have a problem with this change, just curious if it was an errant inclusion since everything else in this PR is API only,

wreality (changes requested):

backend/tests/Feature/UserQueryTest.php
gmeben: I do like your simplification. I restored the assertions for the basic user data, however, because I don't think we should overlook these.

backend/graphql/schema.graphql
```diff
@@ -33,6 +33,50 @@ input ValidateNewUserInput @validator(class: "UserInputValidator") {
...
+"""
+input UpdateProfileMetadataSocialMediaInput @validator(class: "ProfileMetadataSocialMediaValidator") {
+    google: String
+    twitter: String
+    facebook: String
+    instagram: String
+    linkedin: String
+    academia_edu_id: String
+    skype: String
+    zoom: String
```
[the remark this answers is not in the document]
gmeben: I agree. I don't see this being a social media platform and Zoom personal meeting IDs seems to be meant for private sharing: https://support.zoom.us/hc/en-us/articles/201362843-Personal-meeting-ID-PMI-and-personal-link

backend/graphql/schema.graphql
```diff
@@ -33,6 +33,50 @@ input ValidateNewUserInput @validator(class: "UserInputValidator") {
...
+    orchid_id: String
+    humanities_commons: String
+    social_media: UpdateProfileMetadataSocialMediaInput
+    professional_title: String
+    specialization: String
+    affiliation: String
+    interest_keywords: [String]
+    disinterest_keywords: [String]
+    biography: String
+    profile_picture: String
```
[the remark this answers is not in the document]
gmeben: I disagree, but I'm also unsure of what data this field _should_ contain (image data or hyperlinks to images)

backend/graphql/schema.graphql
gmeben: Added.

client/src/layouts/MainLayout.vue
gmeben: Yes, I mentioned this in the PR description.

gmeben: > Bonus: adds a link to the Account Profile page to the sidebar menu on the frontend.

wreality: Oops sorry! I did read it but must have completely forgotten by the time I got to commenting.

backend/app/GraphQL/Validators/ProfileMetadataSocialMediaValidator.php
gmeben: I'll try it

backend/graphql/schema.graphql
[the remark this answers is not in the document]
wreality: I don't personally like storing image data in the dB. Past me has done it for avatars, but I'm now of the opinion that either 
a) an url pattern for avatars as static assets or 
b) the application returning image data read from a file 

Is a much better performing solution. 

I'm imagining an upload/crop/etc your avatar interface. We might want to persist something in the DB about that (a hash or path or something), but that data wouldn't be entered by the user. Everything else in the profile metadata fields are directly entered by the user.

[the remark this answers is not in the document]
gmeben: That all makes sense.

backend/graphql/schema.graphql
```diff
@@ -106,6 +153,42 @@ type User {
     email_verified_at: DateTime
     roles: [Role!]! @belongsToMany
     permissions: [Permission!] @belongsToMany
+    profile_metadata: ProfileMetadata
+}
+
+type ProfileMetadataSocialMedia {
```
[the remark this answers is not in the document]
gmeben: I did this to the types but inputs are still troubling because renaming `UpdateProfileMetadataMailingAddressInput` to `UpdateProfileMetadata_MailingAddressInput` has ramifications that affect naming convention throughout. For example, this can mean that `UpdateProfileMetadataInput` is renamed to `UpdateProfileMetadata_Input` to be consistent, which feels awkward.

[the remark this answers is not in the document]
wreality: This is a good point.. I'm going to let that rattle around and maybe look at some other graphql apis to see what they do.

backend/graphql/schema.graphql
gmeben: No, it doesn't seem to be happily accepting this namespace.

backend/app/GraphQL/Validators/ProfileMetadataSocialMediaValidator.php
gmeben: I'm not able to get this working.

backend/graphql/schema.graphql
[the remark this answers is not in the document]
gmeben: I've implemented a decent compromise: since this feature is destined to be scrapped and re-done later, and because we don't anticipate these types interfering with features elsewhere in the system, why try to navigate around a threat of complexity in the future that's unlikely and theoretical? Why not just simplify the names? We have an opportunity to keep this simple and I think we should take advantage of that.

* `UpdateProfileMetadata_MailingAddressInput` => `UpdateMailingAddressInput`
* `UpdateProfileMetadataSocialMediaInput` => `UpdateSocialMediaInput`
* `ProfileMetadata_SocialMedia` => `SocialMedia`
* `ProfileMetadata_MailingAddress` => `MailingAddress`

We'll keep the validators as they are since we're having issues getting namespaces working with a nested directory structure at `/backend/app/GraphQL/Validators/ProfileMetadata/`.

What do you think?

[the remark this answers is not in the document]
wreality: Love it.  I'll approve

wreality (approved):

Issue #170, opened 2021-03-12
Avatar Color Variants


gmeben:
### Problem/Why
The current default avatar lacks color and does not reflect the potential variety of colors expressed by users in their avatars within the application.

### Goal/What
Create duplicates of the existing avatar for different colors and render those examples in the avatar block.

### Solution/How
Modify the existing source file and create new source files based on this file in the `/client/src/assets` directory and export these as PNGS for different colors to the `/client/public` directory.  Modify the avatarimage component to render a color variation based on email.

### To Do / Acceptance Criteria

- [x] Create assets and export PNGs
- [x] Update AvatarImage to render color avatars

### Review Checklist

- [x] There is a sufficient variety of color variants of the original black and white avatar
- [x] Color variations are deterministically rendered based on email

wreality:
@gmeben @sattler9 @Sruthin86 

I love the images, but I had a thought and did some poking around...   Does anyone have opinions on 1,2 or 3?  

I like 3.  It feels like it creates a better representation of what the UI might look like, plus AI-generated faces are pretty badass IMHO.

## 1) Grant's Images (randomized)
![image](https://user-images.githubusercontent.com/463159/111053097-f8a72e00-842e-11eb-9505-3e7d78ab50cf.png)

## 2) 12 Images sampled from https://randomuser.me/photos
> randomuser.me:
> All randomly generated photos were hand picked from the authorized section of UI Faces.
>
> UI Faces (now defunct)
> Unless they're listed in the authorized section, those awesome folks allow their faces to be used on live products.

![image](https://user-images.githubusercontent.com/463159/111053153-56d41100-842f-11eb-844d-bf50c29e6f40.png)

## 3) A selection of 12 random images from https://thispersondoesnotexist.com/
Quoting from someone who had contacted the author:
https://qr.ae/pN0UzJ
> You are free to use it! Works by an AI are actually not copyrightable

![image](https://user-images.githubusercontent.com/463159/111053210-d2ce5900-842f-11eb-80b6-dff09ac5e73c.png)

wreality:
Sorry, one more variation:
Option 3 plus a few samples from https://thiscatdoesnotexist.com https://thisartdoesnotexist.com and https://thishorsedoesnotexist.com for some more variation.

![image](https://user-images.githubusercontent.com/463159/111053395-5dfc1e80-8431-11eb-8ea7-6b74e02fcd8f.png)

gmeben:
[...]

PR #171, opened 2021-03-12
feat(client): Add avatar color variants

gmeben:
This adds duplicates of the existing avatar for different colors.

Closes Issue #170


wreality:
@gmeben I updated the avatarimage component to render color avatars and updated #170 to include this todo.

wreality (approved):

Issue #141, opened 2021-02-19
Add Publications


gmeben:
**Problem/Why**

There is no formal representation of publications within CCR, which are necessary to group users and submissions.

---

**Goal/What**

Implement publications into CCR 

---

**Solution/How**

---

### To Do / Acceptance Criteria

- [ ]

### Review Checklist

- [ ]

### Accessibility Checklist

- [ ]

gmeben:
The scope of this issue has been adjusted according to our direction outlined in the associated pull request: https://github.com/MESH-Research/CCR/pull/187#discussion_r600727030

PR #187, opened 2021-03-22
feat: add publications

gmeben:
This implements Publications into the system and verifies they can be generically assigned and unassigned to users.

Closes Issue #141


backend/tests/Feature/PublicationTest.php
```diff
@@ -0,0 +1,52 @@
...
+
+class PublicationTest extends TestCase
+{
+    use RefreshDatabase;
+
+    public function testPublicationsCanBeAssignedToAndQueriedForAUser()
+    {
+        $publication = Publication::factory()->create();
+        $user = User::factory()->create();
+        $user->publications()->attach($publication->id);
```
wreality, line 19: So, what does this mean in this context?  Shouldn't a user's association to a publication always be through (or somehow related to, pseudocode-speak) a role?

gmeben: Yes, a user's association to a publication is relative to their role. Remember that users can have more than one role. Here's the breakdown:

- Not associated with publications at all: Application Admin
- Associated with publications directly: Publication Admin, Editor
- Associated with publications indirectly: Review Coordinator, Reviewer, Submitter

The publication-to-user association that's being performed here is a generic one. We'll need to integrate the "business logic" I outlined above later when we develop the code that handles role assignments.

gmeben: It's occurred to me that we should re-evaluate the need of having separate publications on an application instance, and if the complexity of implementing separate publications into the system is justified.

gmeben: Capturing from today's discussion:

We will continue implementing separate publications on an application instance, however, it appears that our permission library may not be designed in a way that satisfies our needs. That is, the permission library seems to be based on a 2-dimensional association (user/role) and we need a 3-dimensional association (user/role/model).

For now, we will reduce the scope of this issue to remove the direct linking between publications and users. We will continue to research solutions to address this association issue separately.

Sruthin86 (approved):
Verified the code.

PR #194, opened 2021-03-30
feat: Add user search to GraphQL API

### LINKED ISSUE #175: Implement fuzzy searching for users api
### Problem/Why
As the user list grows, an application admin should be able to search for a user in the list.

### Goal/What
The application admin should be able to filter the users list via a fuzzy search for email, name, and username.

### Solution/How
Update the GraphQL endpoint to allow a search parameter to be added which implements filtering users.  For inspiration: https://dev.to/khalyomede/create-an-easy-fuzzy-loose-search-with-laravel-5-34a4  

- [x] Update the users list endpoint to accept a search parameter
- [x] Implement fuzzy search logic on username, name, and email

### Review Checklist

- [ ] API returns filtered user results

gmeben:
This extends the GraphQL API to search for users by various search terms.

Closes #175 

Example GraphQL Playground query:

```graphql
query SearchUsers {
  userSearch(term:"abcdef", first: 10, page: 1) {
    data {
      id	
      name
      email
      username
    }
  }
}
```

Edit: changed the name of the function argument from 'search' to 'term'

### DIFF
--- a/backend/config/scout.php
+++ b/backend/config/scout.php
@@ -15,7 +15,7 @@ return [
     |
     */
 
-    'driver' => env('SCOUT_DRIVER', 'algolia'),
+    'driver' => env('SCOUT_DRIVER', 'tntsearch'),
 
     /*
     |--------------------------------------------------------------------------
--- a/backend/config/scout.php
+++ b/backend/config/scout.php
@@ -1,0 +1,119 @@
+<?php
+
+return [
+
+    /*
+    |--------------------------------------------------------------------------
+    | Default Search Engine
+    |--------------------------------------------------------------------------
+    |
+    | This option controls the default search connection that gets used while
+    | using Laravel Scout. This connection is used when syncing all models
+    | to the search service. You should adjust this based on your needs.
+    |
+    | Supported: "algolia", "null"
+    |
+    */
+
+    'driver' => env('SCOUT_DRIVER', 'algolia'),
+
+    /*
+    |--------------------------------------------------------------------------
+    | Index Prefix
+    |--------------------------------------------------------------------------
+    |
+    | Here you may specify a prefix that will be applied to all search index
+    | names used by Scout. This prefix may be useful if you have multiple
+    | "tenants" or applications sharing the same search infrastructure.
+    |
+    */
+
+    'prefix' => env('SCOUT_PREFIX', ''),
+
+    /*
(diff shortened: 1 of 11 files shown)


backend/graphql/schema.graphql
```diff
@@ -21,7 +21,11 @@ type Query {
     "Return pre-defined user roles in the application"
     role(id: ID @eq): Role @find
 
+    "Return a permission by ID"
     permission(id: ID @eq): Permission @find
+
```
wreality, line 26: Should this be a separate query or should the search parameter be added to the existing users query?

backend/config/scout.php
```diff
@@ -0,0 +1,133 @@
...
+    | Here you may configure your Algolia settings. Algolia is a cloud hosted
+    | search engine which works great with Scout out of the box. Just plug
+    | in your application ID and admin API key to get started searching.
+    |
+    */
+
+    'algolia' => [
+        'id' => env('ALGOLIA_APP_ID', ''),
+        'secret' => env('ALGOLIA_SECRET', ''),
+    ],
```
wreality, line 117: Does this depend on an algolia API key to operate?  Excuse my lack of knowledge wrt to scout.

wreality (commented):
Just a few questions to make sure I understand what's going on here.

backend/graphql/schema.graphql
gmeben: Good question. I suppose if I had the option, I would add it to the existing users query, but I have not looked into how to combine two different directives (`eq` and `@search`) on the same query.

backend/config/scout.php
gmeben: Nope, this Algolia section is leftover from the default config file. I can remove this section if you like.

backend/graphql/schema.graphql
wreality: users, not user

I think searchUsers is probably a better name than users. Perhaps this should just replace the users endpoint.

backend/config/scout.php
wreality: Nope that's fine, just wanted to make sure I understood what was going on.

backend/graphql/schema.graphql
gmeben: That's a good suggestion. I'll refactor.

wreality (approved):

Issue #179, opened 2021-03-17
Split schema.graphql into an organized collection of schema files.


wreality:
### Problem/Why
The schema.graphql file is growing in complexity/length and it will become more difficult to locate relevant/related sections as it continues to grow.

### Goal/What
Address how the team would like to organize the schema file while it is small and before it reorganizing it becomes a more monumental task.

### Solution/How
Light house allows importing schema files and using the `extend` keyword to merge types.  The proposed schema organization structure:

```
Root (schema.graphql)
--/users.graphql (types related to user management)
--/auth.graphql (types related to user authentication/authorization)
```
The proposed structure represents currently implemented types. As new types are added, they may be added to an existing schema file, or new schema files could be added as additional groupings become advantageous/evident.

https://lighthouse-php.com/2.6/guides/schema-organisation.html#schema-imports

### To Do / Acceptance Criteria

- [x] Create new schema files
- [x] Move appropriate types to each file
- [x] Add extends/imports as needed to stitch schema together.

### Review Checklist

- [x] Current tests continue to pass.
- [x] Schemas are organized.

PR #201, opened 2021-04-14
refactor: Split schema.graphql into separate files

gmeben:
This extracts from `schema.graphql` each type definition into a separate file.

Closes Issue #179


wreality (approved):
I think there are some future refinements to be made, but to your point earlier, I think this is GTG for now.

Issue #203, opened 2021-04-15
Update profile metadata fields


wreality:
### Problem/Why
Following the discussion, in #185 the profile metadata schema should be updated to match the conclusions.

### Goal/What
The following changes should be made to the graphql schema:
1) Drop Mailing Address
2) Drop Phone
3) Drop Salutation
4) Add a academic_profiles field (sibling to social_media)
5) Move HC profile, academia.edu and orchid id into the new academic_profile

### Solution/How
Update the graphql schema with the appropriate changes and update the front end to save/pull the appropriate fields.  **As we're pre-release, a migration path for the metadata fields is not needed.**

### To Do / Acceptance Criteria

- [x] Update graphql schema to reflect desired structure

### Review Checklist

- [x] Schema changes are implemented

wreality:
Removed the front-end related todo items as this is getting done ahead of the initial frontend implementation for simplicity.

PR #208, opened 2021-04-21
refactor: rearrange profile metadata fields

wreality:
Rearrange metadata fields as outlined in #203


gmeben (approved):
My only thought is to consider adding a max length for `specialization` and perhaps other fields related to strings. If you've already considered this and think it's not worth implementing, then don't worry about it.

wreality:
> My only thought is to consider adding a max length for `specialization` and perhaps other fields related to strings. If you've already considered this and think it's not worth implementing, then don't worry about it.

I just moved fields around, I didn't change any of the validation that was already there.  Its looks like specialization has a max len validator of 256 on it... or am I missing something ?

gmeben:
@wreality You're right. I didn't bother to pop open `ProfileMetadataValidator.php` yesterday and see that it's already there. Everything looks good to go.

PR #271, opened 2021-05-20
fix: Vulnerability with postcss

### LINKED ISSUE #270: Dependabot: Vulnerability with postcss at docs/yarn.lock
CVE-2021-23368

moderate severity

Vulnerable versions: >= 7.0.0, < 8.2.10
Patched version: 8.2.10

The package `postcss` from 7.0.0 and before 8.2.10 are vulnerable to Regular Expression Denial of Service (ReDoS) during source map parsing.

gmeben:
This updates Vuepress dependencies to their latest versions to resolve an outstanding Dependabot alert with postcss: https://github.com/MESH-Research/CCR/security/dependabot/docs/yarn.lock/postcss/open

Dependency update to version 1.8.2:
* https://github.com/vuejs/vuepress/tags
* https://github.com/vuejs/vuepress/blob/master/packages/%40vuepress/plugin-back-to-top/package.json

Closes #270

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -10,8 +10,8 @@
         "build": "vuepress build ."
     },
     "devDependencies": {
-        "@vuepress/plugin-back-to-top": "^1.8.0",
-        "vuepress": "^1.5.3",
+        "@vuepress/plugin-back-to-top": "^1.8.2",
+        "vuepress": "^1.8.2",
         "vuepress-plugin-robots": "^1.0.1",
         "vuepress-plugin-zooming": "^1.1.8"
     }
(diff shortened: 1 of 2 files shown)


gmeben:
The problem with this is that I'm unable to identify how to verify that this will fix the vulnerability in question. The alert says that vuepress requires postcss through various versions of ^7* via (by my count) 42 transitive dependencies.

This may be one of those issues that resolve themselves as package maintainers update their packages, but it doesn't hurt to upgrade our version of vuepress at docs/package.json either.

gmeben:
If it's decided that this is an unnecessary change, we can always dismiss the alert here:

Dialog for dismissing Dependabot alert

I think "risk is tolerable to this project" is an acceptable dismissal.

gmeben:
The team and I decided to close this pull request and dismiss as I recommended.

Issue #174, opened 2021-03-17
Application Admins should be able to view user details


wreality:
### Problem/Why
Application admins can view the list of all users in the system but cannot view more details on each user.

### Goal/What
Application Admins should be able to view details on each user in the system.

### Solution/How
Implement a user details page that displays the details of each user listed on the admin/users index page. 

Use Quasar components and styling to achieve different layouts under different breakpoints (large, medium, small, extra small) according to the prototypes from Issue #196.

### To Do / Acceptance Criteria

- [x] Add a route and page for `admin/users/{id}`
- [x] Implement user fetching from backend
- [x] Display basic user information (username, name, email)
- [x] Adjust page layout based on breakpoints (large, medium, small, extra small)

### Review Checklist

- [ ] Application Admins can view details of each user

### Accessibility Checklist

- [ ] Accessibility check pass

Sruthin86:
Link to the wireframes: https://xd.adobe.com/view/4f17b02d-21d3-42b1-9f5e-fe8262316b1c-1a44/

Sruthin86:
Work completed:

- Implemented user details and tabs for roles on the desktop size 
- Added the correct icons for the user metadata

Work remaining:
- Needs mobile and tablet layout updates.
- Needs conditions to show appropriate data when metadata is not present.

gmeben:
Sections, Icon Set 2 - https://xd.adobe.com/view/398d5b99-9239-485c-ad9f-ff3ed5d42dfc-6cde/

gmeben:
I made progress on this enough to consider it acceptably done according to the criteria. The user details page shows the name, username, email, and (now) role of users. Here's some things to note:

* The listed user items on the All Users page now link to the user details pages.
* I've decided to implement the simpler [section based layout from my designs in XD](https://xd.adobe.com/view/398d5b99-9239-485c-ad9f-ff3ed5d42dfc-6cde/) as my previous accessibility testing of the default QTabs component revealed that this violated the a11y specifications of axe-core 4.2 I have not yet tested QTabs with the recent release of 4.2.1 so this might've changed.
* The relationship that users have to publications and submissions does not yet consider role, so roles can't yet express the publication they're associated with in the roles section.
[...]

PR #289, opened 2021-05-27
feat: Admin-Facing User Details Page

gmeben:
This PR:

* Enables the 'outlined' variant of the material design icon font library
* Fixes the border radius of "square" type avatars
* Adds an admin-only route for viewing user details (`/admin/users/<id>`)
* Makes user items on the All Users page link to this new user details page
* Displays the following data on the user details page:
    - rounded avatar
    - name
    - username
    - email
    - (bonus) roles
* Implements a responsive design layout on the user details page
* Adds a unit test and an integration test for the user details page

Closes Issue #174


client/src/pages/Admin/UserDetails.vue
```diff
@@ -0,0 +1,152 @@
...
+            Roles
+          </div>
+          <div class="col">
+            <div
+              v-for="role in user.roles"
+              :key="role.id"
+              class="text-weight-medium"
+            >
+              <q-icon
+                v-if="role.name === 'Application Administrator' || role.name === 'Publication Administrator'"
```
wreality, line 97: I wonder if we should add a token field to roles table to return a 'token' like APPLICATION_ADMIN with the roles for programmatic matching like this rather than matching a human string that may change in the future if the name of a role changes. 

Not important for this issue but something to think about.

client/src/pages/Admin/UserDetails.vue
```diff
@@ -0,0 +1,152 @@
...
+              </span>
+            </div>
+          </div>
+        </div>
+        <div class="row q-mt-lg">
+          <div class="col-3 q-pr-lg text-right text--grey">
+            Roles
+          </div>
+          <div class="col">
+            <div
```
wreality, line 91: Wrap in a v-if="roles.length"

client/src/pages/Admin/UserDetails.vue
```diff
@@ -0,0 +1,152 @@
...
+                v-if="role.name === 'Application Administrator' || role.name === 'Publication Administrator'"
+                name="manage_accounts"
+              />
+              <q-icon
+                v-if="role.name === 'Editor'"
+                name="o_book"
+              />
+              {{ role.name }}
+            </div>
+            <div v-if="user.roles.length <= 0">
```
wreality, line 106: Make this a v-else on the same level as the v-if="roles.length"

client/src/pages/Admin/UserDetails.vue
```diff
@@ -0,0 +1,152 @@
...
+        />
+        <q-breadcrumbs-el label="User Details" />
+      </q-breadcrumbs>
+    </nav>
+    <div class="row justify-center q-px-lg">
+      <h2
+        class="col-sm-12"
+        data-cy="userDetailsHeading"
+      >
+        {{ user.username || `Loading...` }}
```
wreality, line 17: This would be better handled with a containing element with v-if that checked the graphql loading status. 

https://apollo.vuejs.org/guide/apollo/queries.html#loading-state

client/src/pages/Admin/UserDetails.jest.spec.js
```diff
@@ -0,0 +1,61 @@
...
+        .vm
+        .$options
+        .apollo
+        .user
+        .variables
+        .bind(wrapper.vm)().id
+    ).toBe(1)
+  });
+
+  //! to test how the view handles the data returned, use wrapper.setData(user: {...})
```
wreality, line 59: Should populate the user roles field and test that it renders a list / placeholder if the list is empty. 

(And any other conditional logic in the template, but I think this is the big one)

wreality (changes requested):

client/src/pages/Admin/UserDetails.vue
gmeben: 100%

client/src/pages/Admin/UserDetails.vue
gmeben: Resolved

client/src/pages/Admin/UserDetails.vue
gmeben: Resolved

client/src/pages/Admin/UserDetails.vue
gmeben: Resolved

client/src/pages/Admin/UserDetails.jest.spec.js
gmeben: I'm struggling to get data mocked for this for now, but I'll keep trying. Unless [the documentation](https://vue-test-utils.vuejs.org/guides/getting-started.html) I'm reading is the wrong documentation, I'm finding it to be a bit frustrating because it recommends using methods that are deprecated, like `find()`.

gmeben: It turns out the way I was mocking data was fine, but the assertion wasn't working. I ended up utilizing `wrapper.text()` to check for conditional text being present under different circumstances. The unit test has been extended to check for the following conditional logic in the template:

* A user with no roles
* An application administrator
* A user with no name

wreality (approved):
Added translations to the view and updated the tests.

PR #313, opened 2021-06-08
feat: Add visibility flag to publications

Sruthin86:
**Review checklist**
-  Page only displays the publicly available publications. (As an admin user you should be able to view all publications but as a regular user you should be able to view only publications that are marked as visible in the database) 
-  A menu item linking to the page exits in the sidebar navigation.(Since we do not have access restrictions using the publicly visible flag implemented holding this off for now )
-  Accessibility and integration tests are written.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,6 +30,7 @@ clover.xml
 nbproject/*
 # Visual Studio Code
 .vscode
+*.code-workspace
 # Sass preprocessor
 .sass-cache/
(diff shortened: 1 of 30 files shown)


gmeben (approved):
Things that still need to be done on a later iteration:

* show publicly visible publications for non-admins as opposed to nothing
* create navigation items to more easily access the new "/publications" page

PR #392, opened 2021-07-06
build: Issue368 dependabot javascript

amareshjoshi:


gmeben (approved):
I successfully rebuilt and tested this locally to verify.

PR #432, opened 2021-07-15
test(client): disable animations during tests

wreality:
This PR disables animations during cypress tests using the prefers-reduced-motion media query.  As a side effect this also causes the application to respect prefers-reduced-motion in general.

### DIFF
--- a/client/test/cypress/plugins/index.js
+++ b/client/test/cypress/plugins/index.js
@@ -36,4 +36,15 @@ module.exports = (on, config) => {
       });
     }
   });
+
+  on('before:browser:launch' , (browser = {}, launchOptions) => {
+    const REDUCE = 1;
+    if (browser.family === 'firefox') {
+      launchOptions.preferences['ui.prefersReducedMotion'] = REDUCE;
+    }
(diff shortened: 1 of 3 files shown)


wreality:
@gmeben Take a look.  This should help a lot.

client/test/cypress/plugins/index.js
```diff
@@ -36,4 +36,15 @@ module.exports = (on, config) => {
       });
     }
   });
+
+  on('before:browser:launch' , (browser = {}, launchOptions) => {
+    const REDUCE = 1;
+    if (browser.family === 'firefox') {
+      launchOptions.preferences['ui.prefersReducedMotion'] = REDUCE;
```
gmeben, line 43: I'm curious about the choice to use a local variable here.

gmeben (approved):

client/test/cypress/plugins/index.js
wreality: It was in the example code. I just kept it. 🤷‍♀️

I thought the same thing.

PR #434, opened 2021-07-15
test(cypress): Upgrade cypress

wreality:
Upgrade cypress to the latest version and rearrange tooling so that cypress is installed as a dev yarn dependency.

Advantages:
  - Latest version
  - Cypress version is managed in the package.json, not in random .lando files.
  - Smaller Docker image for Lando to D/l and build
  - Easier to run cypress outside of docker and use the GUI (just run `yarn cypress open` in the client directory)

Caveats:
  - After a lando rebuild, one will need to run `lando cypress install` to install the cypress binaries.
  - **If you currently have cypress enabled in lando extras, disable and reenable it to pull in the new config**

### DIFF
--- a/.github/workflows/test-e2e.yml
+++ b/.github/workflows/test-e2e.yml
@@ -33,7 +33,7 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-node-
       - name: Enable Cypress config
-        run: lando extras e cypress && lando start && lando cypress install
+        run: lando start && lando extras e cypress -y && lando start && lando cypress install
       - name: Migrate Database
         run: lando artisan migrate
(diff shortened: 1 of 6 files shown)


gmeben (approved):
I haven't run this locally to verify anything yet but the changes look good.

I'm surprised there's no way to have Lando automatically run auxiliary commands after running a rebuild.

wreality:
With lando originating images you can, Theres a build option. But when you're using a custom image like cypress the build option doesn't do nuffin. 

This does mean that the cypress executables aren't downloaded with every rebuild too, which could be both an annoyance and a convenience depending on the situation.

PR #436, opened 2021-07-16
style(client): Fix eslint config / enforce stricter rules.

wreality:


wreality:
Closing in favor of #484

PR #483, opened 2021-08-03
feat(api): Add gate directive

wreality:
Adds the `@gate` directive for checking permissions for gates that are not associated with a model.  (For gates with an associated model, use the `@can` directive.


gmeben (commented):
1. ) Should there be a PHPUnit test to generically verify that the `@gate` directive works as expected? 
2. Or is making a generic test not possible because `@gate` must be coupled with an explicit rule and the creation thereof is outside the scope of this work?

PR #690, opened 2021-11-02
test: Increase reliability of SubmissionTest

### LINKED ISSUE #679: Fix PHPUnit Test for Many to Many Relationship
### Problem/Why
Currently, the second test in `SubmissionTest.php` called `testSubmissionsHaveAManyToManyRelationshipWithUsers()` relies on random user assignment. However, there is a part of this test that can cause duplicate user assignment which results in a failed test run as documented here: https://github.com/MESH-Research/CCR/runs/4028576002?check_suite_focus=true

### Goal/What
Fix this test so that duplicate user assignments are no longer possible.

### Solution/How
Use the same approach as `testPublicationsHaveAManyToManyRelationshipWithUsers()` in `PublicationTest` and deliberately select a random user that isn't the same user that was already previously assigned before.

- [x] Fix PHPUnit test code

### Review Checklist

- [x] All existing tests pass

gmeben:
This PR addresses resolving an integrity constraint error that is occasionally produced in the the many-to-many relationship assertion in `SubmissionTest.php`.

Closes #679

### DIFF
--- a/backend/tests/Feature/SubmissionTest.php
+++ b/backend/tests/Feature/SubmissionTest.php
@@ -11,7 +11,6 @@ use App\Models\User;
 use Illuminate\Database\QueryException;
 use Illuminate\Foundation\Testing\RefreshDatabase;
 use Illuminate\Http\UploadedFile;
-use Illuminate\Support\Facades\Storage;
 use Nuwave\Lighthouse\Testing\MakesGraphQLRequests;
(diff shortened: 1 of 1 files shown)


wreality (approved):
I'm not in /love/ with the use of random in a test, but this does indeed solve the problem pointed out.

PR #713, opened 2021-11-17
feat(client): Add notification UI components

### LINKED ISSUE #319: Create wireframes for In-App Notifications
### Problem/Why
Users do not have a way to interact with notifications.

### Goal/What
The interface should:
  - indicate if the user has pending notifications, 
  - allow them to view the pending notifications
  - Dismiss/mark-as-read the notifications.
  - View previous read notifications

### Solution/How
Create/Edit an in browser prototype to show how the user interface should be modified to include notifications.  Leverage existing quasar UI elements as much as possible / practical.  Research existing notification strategies and interfaces.

- [x] Wireframe is approved by the CCR team
- [x] Design follows best practices for accessibility, UI/UX.
- [x] Design patterns flow with existing design patterns

wreality:
Closes #319


client/src/i18n/en-us/index.js
```diff
@@ -207,4 +208,33 @@ export default {
...
+    },
+    submission: {
+      new: {
+        short:
+          "{user.username} has submitted {object.name} to {object.publication.name}",
+      },
+      approved: {
+        short: "{object.name} has been approved",
+      },
+      resubmission: {
```
gmeben, line 235: We should be mindful of and avoid mixing past verbs (`approved`) and nouns (`resubmission`) in our APIs.

Choose either nouns: 

```
request
reply
creation
approval
resubmission
```

or past verbs:

```
requested
replied
created
approved
resubmitted
```

If it were up to me, I would choose nouns because some languages don't have past tense.

gmeben (changes requested):

client/src/components/atoms/NotificationListItem.vue
```diff
@@ -0,0 +1,86 @@
...
+<script>
+import { computed } from "@vue/composition-api"
+import { Screen } from "quasar"
+import { flatten } from "flat"
+import TimeAgo from "javascript-time-ago"
+import en from "javascript-time-ago/locale/en.json"
+
+TimeAgo.addDefaultLocale(en)
+
+const timeAgo = new TimeAgo("en-US")
```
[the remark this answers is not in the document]
wreality: Its not a timezone, its a language locale.  This will likely need to get hooked into the translation API at some point for locale switching, but I'd put that low on the list.

client/src/i18n/en-us/index.js
wreality: Since they're keys and not part of the translation I'm going to lean towards past verb on this bc in the notification sense, the action has happened. I'll push an update.

wreality:
@gmeben This should be ready for a re-review.

client/src/i18n/en-us/index.js
```diff
@@ -207,4 +208,33 @@ export default {
   },
   general_failure: "Oops, there was an error.",
   loading: "Loading...",
+  /*********************
+   * Notifications follow the format of:
+   * <type>: {
+   *    short: "",
+   *    long: "",
```
gmeben, line 215: I notice there's no `long` properties included in the objects here. Is `long` still relevant?

client/src/components/molecules/NotificationPopup.vue
```diff
@@ -0,0 +1,95 @@
...
+  overflow: hidden
+  .q-list.notifications-list
+    display: flex
+    flex-flow: column wrap
+    overflow: hidden
+    .q-item
+      min-height: 40px
+      padding: 2px 16px 2px 0px
+      width: 100%
+      margin-right: 10px
```
gmeben, line 94: I'll take a deeper look at this tomorrow to see if all the styles here are necessary.

client/src/i18n/en-us/index.js
wreality: For now, yes. I anticipate different translation strings for the notifications page vs the pop up.

client/src/components/molecules/NotificationPopup.vue
```diff
@@ -0,0 +1,95 @@
...
+    return { items: notificationItems, isVisible, popupProxy }
+  },
+})
+</script>
+
+<style lang="sass">
+.q-dialog__inner .notifications-container .q-list.notifications-list
+  width: auto
+  .q-item
+    width: 100%
```
gmeben, line 75:
```suggestion
```

These can be safely removed. The styles for `.q-item` below duplicate the `width: 100%`.

client/src/components/molecules/NotificationPopup.vue
```diff
@@ -0,0 +1,95 @@
...
+
+#notifications-wrapper.q-menu
+  display: flex
+  overflow: hidden
+  flex-flow: column
```
gmeben, line 80:
```suggestion
```
These can be safely removed.

client/src/components/molecules/NotificationPopup.vue
```diff
@@ -0,0 +1,95 @@
...
+.notifications-container
+  background: white
+  display: flex
+  flex-flow: column
+  overflow: hidden
```
gmeben, line 85:
```suggestion
```
This can be safely removed.

client/src/components/molecules/NotificationPopup.vue
```diff
@@ -0,0 +1,95 @@
+<template>
+  <q-btn
+    flat
+    padding="none"
+    data-cy="dropdown_notificiations"
+    class="notifications-dropdown"
+    :aria-label="$t('header.notification_button')"
+  >
```
gmeben, line 8: I noticed this introduces an "Interactive controls must not be nested" error in axe DevTools. How should we proceed?

client/src/components/atoms/NotificationListItem.vue
```diff
@@ -0,0 +1,86 @@
+<template>
+  <q-item clickable :class="{ unread: !note.viewed }" class="q-pl-none">
+    <q-badge v-if="!note.viewed" />
+    <q-item-section side class="q-px-md">
+      <q-icon :size="iconSize" :name="note.icon" />
```
gmeben, line 5: I've noticed that the color of these icons does not have a sufficient color contrast when they're against the blue background highlight signifying the notification is unread. Consider making these darker when they're for unread notifications?

client/src/components/molecules/NotificationPopup.vue
```diff
@@ -0,0 +1,95 @@
...
+  display: flex
+  flex-flow: column
+  overflow: hidden
+  .q-list.notifications-list
+    display: flex
+    flex-flow: column wrap
+    overflow: hidden
+    .q-item
+      min-height: 40px
+      padding: 2px 16px 2px 0px
```
gmeben, line 92:
```suggestion
```
I'm kind of a fan of the notification items being taller and noticed that these styles shorten them, which makes the notification text as a whole busier and less easy to read. What do you think?

gmeben (changes requested):

client/src/components/molecules/NotificationPopup.vue
wreality: These are needed for the q-menu version to flow correctly at when the height of the popup can't contain all the items.

client/src/components/molecules/NotificationPopup.vue
wreality: Removing this also breaks the q-menu version of the popup.

client/src/components/molecules/NotificationPopup.vue
wreality: I'll make them slightly taller.

client/src/components/molecules/NotificationPopup.vue
gmeben: I remember that behavior seemed to be preserved when I tested it after removing these.

client/src/components/atoms/NotificationListItem.vue
wreality: This is a size issue and not a color contrast issue.  Fixed.

client/src/components/molecules/NotificationPopup.vue
gmeben: We looked at it and found that the these same styles for the `#notifications-wrapper.q-menu` selector were redundant, but these lines are necessary, so the former were removed.

gmeben (approved):

PR #785, opened 2022-01-14
feat: Add Submission Statuses

### LINKED ISSUE #755: Submission Statuses
### Problem/Why
Currently, there is no established way to set or view the status of a submission. This is important because reviews will rely on this heavily in order for those participating in the review to know where the submission is in its [workflow](https://miro.com/app/board/o9J_l654lxo=/). 

### Goal/What
Implement the backend infrastructure that allows the state of a submission to be designated under the following:

* Initially Submitted
* Awaiting Resubmission
* Resubmitted
* Awaiting Review
* Rejected
* Accepted as Final
* Expired
* Under Review
* Awaiting Decision
* Awaiting Revision
* Archived
* Deleted

### Solution/How
Write a migration to update submission table with a new column `status`.  Update the GraphQL schema to reflect the new `status` property of submissions using graphql Enums. Write a PHPUnit test to assert that a submission's status can be retrieved and changed via Eloquent.

- [x] Write migration
- [x] Update `Submission` model
- [x] Update GraphQL schema
- [x] Add PHPUnit testing

### Review Checklist

- [ ] The status of a submission can be viewed via query from the GraphQL endpoint
- [ ] The status of a submission can be viewed via Eloquent (ex: `$submission->status`)
- [ ] The status of a submission can be changed via Eloquent
- [ ] All new and existing tests pass

gmeben:
This PR:

* Adds `status` and `status_name` properties to the submissions model
* Adds a migration for the new column in the submissions table
* Bonus: Fixes an emergent bug with ordering (again somehow!) in the `testCreateSubmissionUserViaMutationAsAReviewCoordinator()` method of `SubmissionTest.php`

Closes #755

### DIFF
--- a/backend/app/Models/Submission.php
+++ b/backend/app/Models/Submission.php
@@ -38,6 +38,15 @@ class Submission extends Model
     ];
 
     /**
+     * The accessors to append to the model's array form.
+     *
+     * @var array
+     */
+    protected $appends = [
+        'status_name',
+    ];
+
+    /**
      * The publication that the submission belongs to
      *
      * @return \Illuminate\Database\Eloquent\Relations\BelongsTo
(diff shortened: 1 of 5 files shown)


backend/app/Models/Submission.php
```diff
@@ -54,4 +64,27 @@ public function files(): HasMany
...
         return $this->hasMany(SubmissionFile::class);
     }
+
+    /**
+     * @return string
+     */
+    public function getStatusNameAttribute()
+    {
+        $statuses = [
+            1 => 'Initially Submitted',
```
wreality, line 74: Should these be strings that are easier to match on the front end rather than human readable. Eg. "INITAL_SUBMISSION"

backend/graphql/submission.graphql
```diff
@@ -10,6 +10,8 @@ type Submission {
     users: [User!]! @belongsToMany
     files: [SubmissionFile]! @hasMany
     pivot: SubmissionUser
+    status: Int!
```
wreality, line 13: I strongly suggest adding a graphql enum type for this.

wreality (changes requested):
Suggested some changes to the text for statuses, also suggest using an Enum in the graphql schema.

backend/graphql/submission.graphql
gmeben: This has been refactored as advised. This value is now represented as an enum.

backend/app/Models/Submission.php
gmeben: This has been refactored as advised. These are now underscore-delimited and in all caps.

wreality (approved):

PR #799, opened 2022-01-24
build(deps): upgrade front end to quasarv2/vue3

wreality:
Complete migration to quasarv2 and vue3.


wreality:
@gmeben @Sruthin86  Feel free to start reviewing this.  There's a lot, so we should probably touch base during team time about what a review means for this one.

Also, cypress is failing in CI, but passing locally.  I'll be digging into it a bit, but it seems to be a cypress crash rather than a failing test.

gmeben (approved):

client/src/pages/Admin/PublicationDetails.jest.spec.js
```diff
@@ -1,51 +1,51 @@
-import { mountQuasar } from "@quasar/quasar-app-extension-testing-unit-jest"
+//import { jest, describe, test, it } from "@jest/globals"
```
wreality, line 1:
```suggestion
```

wreality:
Using my admin override to publish this with cypress being funky as that will get add shortly.

PR #813, opened 2022-01-27
feat: Notification Read Statuses

### LINKED ISSUE #700: Handle Notification Read Statuses
### Problem/Why
There is currently no implemented way for in-app notifications to be indicated as read by a user. Users may annoyingly encounter the same notifications when navigating throughout CCR with no way to mark notifications as read and dismiss them.

### Goal/What
Implement a way for users to mark notifications as read.

### Solution/How
Define user interaction behavior for marking notifications as read. Research strategies for implementing read status demarcation (https://laravel.com/docs/8.x/notifications#marking-notifications-as-read) via user interaction, using Lighthouse if possible. Update code as necessary on the client and the backend. 

- [x] Create a mutation to update the read status for an individual notification
- [x] Have a guard in place to ensure that only the current user can update their read status
- [x] Create a "mark all read" mutation to update the status of all the notifications
- [x] Add interaction on the UI to enable marking notifications as read
- [x] Add PHPUnit tests (as necessary)
- [x] Add Jest testing to make assertions about notifications being read and unread 
- [x] Add Cypress testing to assert that notifications can be marked as read individually
- [x] Add Cypress testing to assert that multiple notifications can be marked as read simultaneously 

### Review Checklist

- [ ] All new and existing tests pass

gmeben:
This PR:

* Allows users to mark a notification as read
* Adds a second submission to the submission seeded data
* Generates notifications for the submission seeded data

Closes #700


Sruthin86 (approved):

backend/tests/Feature/NotificationTest.php
```diff
@@ -227,4 +227,92 @@ public function testNotificationsCanOnlyBeQueriedForOneself()
...
+        $notification_data_1 = $this->getSampleNotificationData($user, 1012);
+        $user->notify(new SubmissionCreated($notification_data_1));
+        $notification_data_2 = $this->getSampleNotificationData($user, 1013);
+        $user->notify(new SubmissionCreated($notification_data_2));
+        $this->graphQL(
+            'mutation MarkAllNotificationsRead {
+                markAllNotificationsRead
+            }'
+        );
+        $this->assertEquals(0, $user->unreadNotifications()->count());
```
wreality, line 295: Probably should assert the return value of the mutation here as well (count of notifications marked as read)

client/src/components/atoms/NotificationListItem.vue
```diff
@@ -58,17 +66,35 @@ const props = defineProps({
...
-  return flatten(props.note)
+  const flattened = flatten(props.note)
+
+  const mutated_keys = {}
+  Object.keys(flattened).forEach((k) => {
+    // console.log("k", k.replaceAll("data.", "").replaceAll(".", "_"))
+    mutated_keys[k.replaceAll("data.", "").replaceAll(".", "_")] = flattened[k]
+  })
+  console.log("mutated", mutated_keys)
+  return mutated_keys
```
wreality, line 97:
```suggestion
  return flatten(props.note, {delimiter: "_")
```

client/src/components/molecules/NotificationPopup.jest.spec.js
```diff
@@ -28,36 +28,57 @@ describe("Nofitication Popup", () => {
...
+  function getNotificationData(readStatus) {
+    const data = [
+      {
+        data: {
+          id: "df91c4c0-3cf2-409d-a305-9a6005149c86",
+          type: "AppNotificationsSubmissionCreated",
+          notifiable_type: "AppModelsUser",
+          notifiable_id: 1,
+          data: `"{"submission":{"id":9999,"title":"A Modest Proposal"},"publication":{"id":9999,"name":"Test Publication from Tinker"},"user":{"id":9999,"username":"Test User from Tinker"},"type":"submission.created","body":"A submission has been created.","action":"Visit CCR","url":"/"}"`,
+          read_at: null,
```
wreality, line 40:
```suggestion
          read_at: readStatus ? "2021-12-31 12:15:15" : null,
```

client/src/components/molecules/NotificationPopup.jest.spec.js
```diff
@@ -28,36 +28,57 @@ describe("Nofitication Popup", () => {
...
+          created_at: "2021-12-13 20:13:05",
+          updated_at: "2021-12-13 20:13:05",
+        },
       },
-    },
-  ]
+    ]
+    if (readStatus === "read") {
+      data[0]["read_at"] = "2021-12-31 12:15:15"
+    }
```
wreality, line 48:
```suggestion
```

client/src/components/molecules/NotificationPopup.jest.spec.js
```diff
@@ -28,36 +28,57 @@ describe("Nofitication Popup", () => {
...
+    return data
+  }
 
   it("mounts without errors", () => {
     const { mockClient } = wrapperFactory()
-    const queryHandler = jest.fn().mockResolvedValue(notification_data)
+    const queryHandler = jest
+      .fn()
+      .mockResolvedValue(getNotificationData("unread"))
```
wreality, line 56:
```suggestion
      .mockResolvedValue(getNotificationData(false))
```

client/src/components/molecules/NotificationPopup.jest.spec.js
```diff
@@ -28,36 +28,57 @@ describe("Nofitication Popup", () => {
...
     expect(wrapperFactory().wrapper).toBeTruthy()
   })
 
-  it("displays an indicator for a user that has notifications", async () => {
+  it("displays an indicator for a user that has unread notifications", async () => {
     const { wrapper, mockClient } = wrapperFactory()
-    const queryHandler = jest.fn().mockResolvedValue(notification_data)
+    const queryHandler = jest
+      .fn()
+      .mockResolvedValue(getNotificationData("unread"))
```
wreality, line 65:
```suggestion
      .mockResolvedValue(getNotificationData(false))
```

client/src/components/molecules/NotificationPopup.jest.spec.js
```diff
@@ -28,36 +28,57 @@ describe("Nofitication Popup", () => {
...
     mockClient.setRequestHandler(CURRENT_USER_NOTIFICATIONS, queryHandler)
     const indicator = wrapper.findComponent({ ref: "notification_indicator" })
     expect(indicator).toBeTruthy()
   })
 
+  it("displays no indicator for a user that has no unread notifications", async () => {
+    const { wrapper, mockClient } = wrapperFactory()
+    const queryHandler = jest
+      .fn()
+      .mockResolvedValue(getNotificationData("read"))
```
wreality, line 75:
```suggestion
      .mockResolvedValue(getNotificationData(true))
```

wreality (changes requested):
Couple of things to clean up, but overall it looks really well done.

backend/tests/Feature/NotificationTest.php
gmeben: Added

client/src/components/atoms/NotificationListItem.vue
gmeben: That delimiter parameter is nice, however, the problem is that the flattened data is returned with `data_` prepended to every property name. So, `data_` has to be removed in order for the mapping to work. The alternative is to update the translation files by nesting them inside a `data` object and I don't know if that's better.

gmeben: After trying both, I've decided it's better, so this has been refactored.

wreality: I mean you could flatten(props.note.data, ....  no?

gmeben: Then other top-level properties like `id` won't be included.

wreality (approved):

PR #852, opened 2022-02-09
test(client): remove undeeded cyattr prop

wreality:
This PR:

- Removes the cyAttr prop from VQInput ($attrs are bound to the Q-Input so data-cy is passed down automatically)

### DIFF
--- a/client/src/components/atoms/VQInput.vue
+++ b/client/src/components/atoms/VQInput.vue
@@ -7,7 +7,6 @@
     v-bind="$attrs"
     ref="input"
     v-model="model"
-    :data-cy="cyAttr"
     :error="v.$error"
     :label="getTranslation('label')"
     :hint="getTranslation('hint')"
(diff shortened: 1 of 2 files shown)


gmeben (commented):
I wanted to capture what I mentioned earlier today:

I noted back in November 2021 that the `cyAttr` prop was added to this nested QInput component in order to avoid having the attribute applied to both the label and the input of a QInput component. This may or may no longer be a problem. If it's found to no longer be a problem, then these changes should be fine.

wreality:
I guess I’m still failing to see how this would have solved that problem originally but I’ll take your word for it.

I’ll double check that this isn’t the case and either request review or close depending on the outcome.
________________________________

gmeben:
I suspect the error is a quirk of Quasar, similar to how Quasar will add both a label element for a QInput as well as give the QInput's input element an `aria-label` attribute. For some reason, adding the attribute in the traditional, static manner duplicated the attribute on both elements. Adding the attribute via prop didn't. I'll do some experimentation today to see if this is still the case.

gmeben:
I was able to consistently encounter the same error locally on `profile.spec.js` that the test runner encounters. 

The experimentation I performed today revealed that the behavior I described earlier is no longer happening for QInput components, however, I was able to replicate this on the EditableList component by setting `v-bind="$attrs"` on its nested QInput component. The data-cy attribute gets applied to both the parent div as well as the input, which suggests an issue with nesting and binding.

gmeben (approved):

PR #856, opened 2022-02-13
ci(cypress): improve cypress test reliability

wreality:
This PR updates the cypress workfow to:
- Build the UI in a separate job
- Add cypress.io configuation
- Downgrades cypress to version 8.3 due to performance bug in later versions
- Performs E2E tests on a built client


wreality:
So, we still have flaky tests happening (there have been between 4-7 failures of this PR in various runs, without any changes to the test/source code).  There have, however, not been any "Crash" failures, and the test time has been improved by about 5 fold.

gmeben (approved):

Issue #833, opened 2022-02-02
Wireframes for Submission Comment/Review


stephanievaskomsu:
### Problem/Why
There is no established view/feel of what the the submission viewing/editing page looks like. 

### Goal/What
Create wireframes for submission commenting and holistic review on the submission viewing/editing page.

### Solution/How
Develop a mockup/wireframe within Miro.

### To Do / Acceptance Criteria
- [x] Create low-fidelity wireframes for Monday Feb 5

### Review Checklist
- [x] Submission text content is represented within the wireframe
- [x] Holistic comments are represented within the wireframe
- [x] Inline comments are represented within the wireframe

### Accessibility Checklist
- [ ] The layout of the wireframe can be adapted to differently sized viewports

sattler9:
Feedback from PPJ on Oct. 4th, 2021 at the editorial / CCR meeting on this [miro board](https://miro.com/app/board/o9J_ltyWt-w=/)

Kurt likes using different versions for the reader’s report (ver F + B) and PHC  - “for reader report comments (private), i'd vote for variation F (on the review page) but with the comments in the sidebar so folks can read the submission and any in-line comments as they compose those summary comments; for holistic comments (public), it might be interesting/useful to put them on the submission details so you have a mechanism to get folks back into the system”

Shelby likes A or C for PHC

Chris seems to like A or A-1 or C. Also wants the PHC tightly linked but separate.

Whispers more for the reader’s report to speak to the editor.

Sidebar is good for the reader’s report - general agreement by PPJ Ed team

May want to refer to section of the article as they make comments

gmeben:
I seem to have moved this into the Icebox column by accident yesterday. 

I'm going to move this into the In Progress column and assign @stephanievaskomsu since she's already worked on this. If a velocity needs to be associated with this issue, one can be added that includes the effort expended during a typical workday (like we previously discussed as a team on in our review/retrospective meeting on Tuesday).

PR #907, opened 2022-02-25
feat: Mocked Submission View

### LINKED ISSUE #834: Mock Review Submission View with Mocked Data
### Problem/Why
Currently, there's no layout(s) of a page to view a submission's contents within CCR for a user to interact with. 

### Goal/What
Have a simple interface in CCR created with mocked text-based content to represent how a manuscript would appear within CCR to those participating in its review process. This should implement the design from the wireframes created in Issue #833. This view should be accessible from the Submission Details page.

### Solution/How
Design within the browser a prototype page using Quasar components following the design of the wireframes. The text should be static and mocked (lorem ipsum or equivalent). There should be elements to represent the content generated from interactions like holistic and inline commenting. 

- [x] Add hyperlink to the Submission Details page
- [x] Design a page according to the wireframes from #833
- [x] Add a Cypress test asserting that this page can be navigated to via links after login

### Review Checklist

- [x] Submission text is represented
- [x] Holistic comments are represented
- [x] Inline comments are represented

### Accessibility Checklist

- [x] The submission text adapts to changing viewports in a way that remains readable
- [x] There are no color contrast issues introduced
- [x] Submission text is detectable by a screen reader
- [x] Holistic comment text is detectable by a screen reader
- [x] Inline comment text is detectable by a screen reader
- [x] Inline comments can be accessed via keyboard interactions exclusively

gmeben:
This adds a simple interface in CCR created with mocked text-based content to represent how a manuscript would appear within CCR to those participating in its review process. It's based on the wireframes created in Issue #833.


client/src/components/atoms/InlineComment.vue
```diff
@@ -0,0 +1,69 @@
+<template>
```
wreality, line 1: The diff between this file and InlineCommentReply, make me think they could easily be the same component.  Given that multi-root components are a thing now, I'd also consider moving the rendering of comment replies part of this component (and recursively render InlineComment with mayne a "isReply" boolean prop)

client/src/components/atoms/InlineCommentReply.vue
```diff
@@ -0,0 +1,72 @@
+<template>
```
wreality, line 1: See comment on InlineComment

client/src/components/atoms/InlineComment.vue
```diff
@@ -0,0 +1,69 @@
...
+</template>
+
+<script setup>
+import AvatarImage from "./AvatarImage.vue"
+defineProps({
+  user: {
+    type: Object,
+    default: () => {},
+  },
+})
```
wreality, line 68:
```suggestion
const user = {}
```
Since this won't be coming in as a prop (rather "comment" will) I'd make this a var rather than a prop just to avoid confusing ourself when we implement into thinking this needs to be a prop.

client/src/router/routes.js
```diff
@@ -87,6 +87,15 @@ const routes = [
...
         props: true,
       },
+      {
+        name: "submission_view",
+        path: "/submission/view/:id",
+        component: () => import("src/pages/SubmissionView.vue"),
+        meta: {
+          requiresSubmissionAccess: true,
+        },
+        props: true,
```
wreality, line 97:
```suggestion
        name: "submission_review",
        path: "/submission/review/:id",
        component: () => import("src/pages/SubmissionReview.vue"),
        meta: {
          requiresSubmissionAccess: true,
        },
        props: true,
```
Ofc, this would necessitate renaming submissionview.vue as well.

client/src/pages/SubmissionView.vue
```diff
@@ -0,0 +1,507 @@
...
+          <q-btn
+            aria-label="Toggle Inline Comments"
+            dense
+            flat
+            round
+            icon="question_answer"
+            @click="toggleRightDrawer"
+          />
+        </q-toolbar>
+      </q-header>
```
wreality, line 38: I would make this a component, SubmissionReviewToolbar (or header)

client/src/pages/SubmissionView.vue
```diff
@@ -0,0 +1,507 @@
...
+            <div class="q-ml-md q-mb-md">
+              <inline-comment-reply :user="currentUser" />
+              <inline-comment-reply :user="currentUser" />
+            </div>
+            <div class="row justify-center q-pa-md">
+              <q-btn color="dark" icon="arrow_upward">Scroll to Top</q-btn>
+            </div>
+          </q-scroll-area>
+        </div>
+      </q-drawer>
```
wreality, line 64: I'd also make this a component, SubmissionReviewInlineDrawer

client/src/pages/SubmissionView.vue
```diff
@@ -0,0 +1,507 @@
+<template>
+  <div v-if="!submission" class="q-pa-lg">
+    {{ $t("loading") }}
+  </div>
+  <article v-else>
```
wreality, line 5: I'm not sure these are semantically used.  The article tag around the submission content 👍, this one seems to be encapsulating more of the interface/chrome.

client/src/pages/SubmissionView.vue
```diff
@@ -0,0 +1,507 @@
...
+          <p>
+            orci eu lobortis. At quis risus sed vulputate odio ut enim blandit
+            volutpat. Tempor id eu nisl nunc mi. Malesuada nunc vel risus
+            commodo viverra maecenas accumsan lacus. Porttitor leo a diam
+            sollicitudin tempor. Blandit massa enim nec dui nunc mattis enim.
+            Elementum nisi quis eleifend quam adipiscing vitae proin sagittis.
+            Placerat orci nulla pellentesque dignissim. Condimentum id venenatis
+            a condimentum vitae sapien pellentesque.
+          </p>
+        </article>
```
wreality, line 156: SubmissionReviewContent  Component

client/src/pages/SubmissionView.vue
```diff
@@ -0,0 +1,507 @@
...
+        <q-separator class="page-seperator" />
+        <section class="comments">
+          <div class="comments-wrapper">
+            <h3 class="text-h1">Overall Comments</h3>
+            <q-card>
```
wreality, line 161: I would do the same thing I was suggesting with inlinecomments (single component, recursive (but single level bc we're not fully nesting comments ofc)

client/src/pages/SubmissionView.vue
```diff
@@ -0,0 +1,507 @@
...
+                            <q-item clickable>
+                              <q-item-section>Send Feedback</q-item-section>
+                            </q-item>
+                            <q-item clickable>
+                              <q-item-section>Share</q-item-section>
+                            </q-item>
+                          </q-list>
+                        </q-menu>
+                      </q-btn>
+                    </div>
```
wreality, line 273: Same deal as in InlineComments.  Im not sure if they can share the same component (i.e. will have the same actions), but that could perhaps be a refactor if the two end up having different context menus.

wreality (changes requested):
Take a look at how I structured the account and profile pages.  The page itself, is a high level wrapper and also where the graphql work happens.  Events are passed to mutate state as needed.  I'd heavily componetize the Page so that a lot of the interaction logic can happen inside the components.

wreality:
Sorry, some of my comments didn't end up getting inside the review, but they're there either way. :-D

client/src/pages/SubmissionView.vue
gmeben: Well, the interface/chrome contains content that constitute this part of the page as an independent, stand-alone piece of the site as defined by the HTML living standard. That standard also includes an example where the `header` is considered a child of its parent `article`: https://html.spec.whatwg.org/multipage/sections.html#the-article-element

This example demonstrates how a `header` and `footer` can be considered semantically part of their parent `article`: https://ilikekillnerds.com/2014/07/how-to-correctly-use-semantic-html5-article-main-and-section-tags/

This example demonstrates how an `aside` element can be considered semantically part of its parent `article`: https://html5doctor.com/aside-revisited/

Our implementation includes a `header` and an `aside` as part of the parent `article`.

client/src/router/routes.js
wreality: Well it's not the submission view so what do you suggest?

gmeben: How is it not the submission view?

client/src/components/atoms/InlineComment.vue
gmeben: This has been refactored as suggested.

client/src/components/atoms/InlineCommentReply.vue
gmeben: This has been refactored as suggested.

client/src/pages/SubmissionView.vue
gmeben: This has been refactored as suggested.

client/src/pages/SubmissionView.vue
gmeben: This has been refactored as suggested.

client/src/pages/SubmissionView.vue
gmeben: This has been refactored as suggested.

client/src/pages/SubmissionView.vue
gmeben: This has been refactored as suggested.

client/src/pages/SubmissionView.vue
gmeben: This has been refactored as suggested.

client/src/router/routes.js
gmeben: Since I recognize this can go either way, I'll rename occurrences of "view" to "review" as suggested.

client/src/components/atoms/InlineComment.vue
```diff
@@ -0,0 +1,48 @@
+<template>
```
wreality, line 1: At a quick glance, it looks likes this isn't referenced anywhere anymore. delete?

wreality (approved):
Approved with a couple of very minor conversations to resolve.  Appreciate the work on this!

client/src/graphql/queries.js
```diff
@@ -81,6 +81,11 @@ export const GET_USERS = gql`
         name
         username
         email
+        pivot {
+          id
+          role_id
+          submission_id
+        }
```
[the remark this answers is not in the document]
gmeben: Yes! Good catch

client/src/components/atoms/InlineCommentReply.vue
```diff
@@ -0,0 +1,56 @@
...
```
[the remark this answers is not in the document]
gmeben: Indeed

PR #938, opened 2022-03-11
fix(client): prevent logout link when dirtyGuard is active

### LINKED ISSUE #840: useDirty guard doesn't work when clicking logout
### Problem/Why
If the user clicks logout while in a form that uses the dirtyguard composable (account profile, metadata) the dialog appears, but the user it logged out before they have an opportunity to choose a response.

### Goal/What
The logout navigation should wait until the dirtyguard has confirmed the user wants to exit the form.

### Solution/How
The logout button doesn't use the router, but instead calls the logout mutation directly.  The dirtyGuard composable users the onRouteBeforeLeave callback, which only is called on a new navigation requested of the router.  The logout link should link to a new page "Logout" that actually process the logout and redirects the user to the home page of the application.

- [x] Create Logout.vue which onMounted triggers the logout actions that are currently called by the logout link
- [x] Update Logout link in header to trigger a router.push to the new logout page.
- [x] Update header cypress test to validate functionality

### Review Checklist

- [x] New tests are complete and passing
- [x] Existing jest tests pass
- [x] Existing cypress tests pass

### Accessibility Checklist
No accessibility issues wrt to this change.

wreality:
Closes #840


client/src/use/forms.jest.spec.js
```diff
@@ -33,18 +33,15 @@ jest.mock("quasar", () => ({
...
 
-    const mockNext = jest.fn()
-    dirtyGuardCallback(null, null, mockNext)
-    expect(mockNext).toHaveBeenCalledTimes(1)
-    expect(mockNext).toHaveBeenCalledWith()
+    expect(await dirtyGuardCallback()).toBe(true)
   })
 
-  test("Shows dialog appropriatly and correctly responds to user feedback", () => {
+  test("Shows dialog appropriatly and correctly responds to user feedback", async () => {
```
gmeben, line 44:
```suggestion
  test("Shows dialog appropriately and correctly responds to user feedback", async () => {
```

gmeben (changes requested):

gmeben (approved):

PR #993, opened 2022-04-07
feat: Optional style criteria validation

gmeben:
This PR:

* Adds a dialog that appears when no style criteria are selected before submitting an inline comment
* Refactors editor buttons into their own atom component `CommentEditorButton`
* Refactors static text into i18n.js-derived text
    * The mocked text for style criteria has not been refactored this way because it's eventually going to be database generated


client/src/components/atoms/CommentEditorButton.vue
```diff
@@ -0,0 +1,50 @@
+<template>
+  <q-btn
+    :aria-label="$t(props.ariaLabel)"
```
wreality, line 3: So, anything that's not a prop will automatically be applied to the root element as an attr via inheritAttrs.  I noticed you're specifying props that are just applied to the root element as is more often lately.  Is there a reason to not take advantage of vue's default behavior that you see here ?

client/src/components/atoms/CommentEditorButton.vue
```diff
@@ -0,0 +1,50 @@
...
+    :class="{ 'is-active': props.isActive }"
+    :disabled="props.isDisabled"
+    @click="props.clickHandler"
+  >
+    <q-tooltip
+      class="bg-primary"
+      anchor="top middle"
+      self="bottom middle"
+      :offset="[0, 5]"
+      >{{ $t(props.tooltipText) }}</q-tooltip
```
wreality, line 17:
```suggestion
      >{{ props.tooltipText }}</q-tooltip
```
I'd strongly suggest sending in translated string rather than translation keys for consistency.  This also removes the dependency on useI18n in the component making it easier to test

I'd recommend using i18n in the parent component and sending in a translated string.

client/src/components/forms/CommentEditor.vue
```diff
@@ -1,205 +1,216 @@
...
 
-export default {
-  components: {
-    EditorContent,
+const commentEditorButtons = ref([
+  {
+    ariaLabel: "guiElements.button.bold.ariaLabel",
+    isActive: computed(() => editor.value.isActive("bold")),
+    clickHandler: () => editor.value.chain().focus().toggleBold().run(),
+    tooltipText: "guiElements.button.bold.tooltipText",
```
wreality, line 77:
```suggestion
    tooltipText: t("guiElements.button.bold.tooltipText"),
```
related to my earlier comment

wreality (changes requested):

client/src/components/atoms/CommentEditorButton.vue
gmeben: I placed this in the child rather than the parent because it made the code more DRY. 

I'll refactor as suggested.

wreality: Yeah, I can totally appreciate that. It’s a give and take between dependencies and dry in this case probably. 

Another thought, if we’re going to do this, we should perhaps use a specifically named prop (tAriaLabel maybe?) for future clarity’s sake?

gmeben: This has been refactored.

client/src/components/forms/CommentEditor.vue
gmeben: This has been refactored.

wreality (approved):

PR #1009, opened 2022-04-19
build(deps,client): switch/upgrade to new @quasar/app-webpack package

wreality:
Per https://github.com/quasarframework/quasar/releases/tag/%40quasar%2Fapp-v3.4.0

Quasar switched from using @quasar/app to @quasar/app-webpack for the application dep.  This PR upgrades to that new package and resolves an outstanding vuln.


Sruthin86 (approved):
We looked at it during standup and approving this change.

PR #1019, opened 2022-04-22
build(lando): add pandoc to appserver lando container

wreality:
Adds pandoc to the appserver container.  Be sure to rebuld in order to install pandoc.  `lando rebuild -y`

This PR also adds pandoc to the lando tooling commands allowing pandoc usage via lando commands:
`lando pandoc -v`

image


gmeben (approved):
Excellent work. I'm glad this could be done under Lando's tooling instead of as a service. 

I believe this necessitates a corresponding [docs](https://ccr.meshresearch.dev) update, but I'm approving this in the meantime.

PR #1067, opened 2022-05-03
refactor(client,api): Split submission queries and mutations into separate fields

wreality:
* Refactors submissions queries and mutations into individual fields by role
* Refactors front end to use the refactored API
* Refactors submission policies to improve readability
* Adds new argPolicy directive to apply policies to individual input arguments
* Adds createsubmissionfiles directive to save files while using lighthouse @create directive rather than custom muation resolver
* Applies policies correctly to prevent unrelated users from viewing submission
* Changes unique constraint on submission_user to prevent multiple role assignment per submission
* Refactors PHPUnit tests to remove redundant tests
* Updates front end to limit review_coordinators to a single user


backend/app/Policies/SubmissionPolicy.php
```diff
@@ -0,0 +1,141 @@
...
+     * @param \App\Models\Submission $submission
+     * @return bool|\Illuminate\Auth\Access\Response
+     */
+    public function updateSubmitters(User $user, Submission $submission)
+    {
+        if ($this->checkAdminRoles($user, $submission->publication_id)) {
+            return true;
+        }
+
+        //Check is the user is a submitter or review coordinator
```
wreality, line 55:
```suggestion
        //Check if the user is a submitter or review coordinator
```

wreality:
@gmeben pretty sure I figured out the right refactor for the comments mutations but this is already a huge PR. I'll refactor the comment mutations in a separate pr / issue. 

Should otherwise be ready for review.

gmeben (approved):
I'm not able to get `testThatAUserCanBeSearchedBySearchTerms()` passing locally in `UserQueryTest.php` for the first three provided test cases. I also have 2 out of the 54 Cypress tests failing: `should allow assignments of editors and reject assignments of duplicate editors` and `should disallow assignments of duplicate reviewers`. 

I must be missing something locally. Other than that, the code looks good.

wreality:
I get the same problem on my local with those user search tests. I haven't gotten to the bottom of why yet. Need to dig in further. 

Can you add what the failure point for those other tests is? It's probably find user select again, but just in case I'd like to dig in a little and check.

PR #1078, opened 2022-05-05
fix(client): Fix lockup in find-user-select component

### LINKED ISSUE #1056: FindUserSelect component locks up occasionally.
### Problem/Why
The find-user-select component (and ultimately the QSelect component) occasionally locks up on user input.

### Goal/What
The component should display the list of matching users and not get stuck in a loading state.

### Solution/How
Changing debounce value has not seemed to help.  Likely this is related to mutations getting cancelled during typing or a need for a filter-abort handler on the QSelect.  Investigation is required to determine the root issue and solve it.  If unsolvable using QSelect, we should implement our own input component that behaves correctly.

- [ ] Determine root cause of lockup
- [ ] Implement fix for component
- [ ] Rejoice.

### Review Checklist

- [ ] QSelect no longer locks up

### Accessibility Checklist
If a new component is used, it should pass accessibility automated tests AND be accessible.

wreality:
Updates find user select to work around graphql/apollo locking up on quick queries. Bypasses the update mechanism from quasar to instead reactively update the query variables and call refetch on the original query which allows vueapollo to cancel duplicate queries on its own.

Once this gets merged it *should* clear up the two apollo dependabot PRs chillin in the open list.

Closes #1056.

### DIFF
--- a/client/test/cypress/integration/submissiondetails.spec.js
+++ b/client/test/cypress/integration/submissiondetails.spec.js
@@ -21,43 +21,27 @@ describe("Submissions Details", () => {
     cy.dataCy("input_review_assignee").type("applicationAd")
     cy.dataCy("result_review_assignee").click()
     cy.dataCy("review_assignee_selected").contains("applicationAdminUser")
-    cy.dataCy("input_review_assignee").type(
-      "test{backspace}{backspace}{backspace}{backspace}"
-    )
+
     cy.dataCy("button_assign_reviewer").click()
     cy.dataCy("list_assigned_reviewers").contains("Application Administrator")
     cy.injectAxe()
     cy.dataCy("submission_details_notify")
 
-    cy.checkA11y(null, {
-      rules: {
-        "nested-interactive": { enabled: false },
-      },
-    }, a11yLogViolations)
+    cy.checkA11y(null, null, a11yLogViolations)
   })
 
   it("should allow assignments of reviewers by review coordinators", () => {
     cy.task("resetDb")
     cy.login({ email: "reviewcoordinator@ccrproject.dev" })
     cy.visit("submission/100")
-    cy.dataCy("input_review_assignee").type("applicationAd{backspace}{backspace}")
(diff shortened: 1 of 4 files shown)


gmeben (approved):
I was unable to successfully run the `submissiondetails.spec.js` locally after five attempts wherein 4 out of the 5 tests consistently failed for the same reasons. The first 3 of the failed tests resulted from color contrast issues related to the notify element, and the last failed test resulted because the notify element representing the error message never appears. 

I'm not sure why origin passes and mine does not. I rebuilt and also separately updated my client dependencies with `lando yarn upgrade` which listed `axe-core@4.4.1` and `cypress-axe@0.14.0` Running the command `yarn list --pattern axe` locally shows `axe-core@4.4.1` and `cypress-axe@0.14.0`.

I'll approve since it consistently passed on all attempts on origin. There must be something wrong with my local setup.

gmeben:
The only questionable thing I noticed was the removal of the exception to ignore "nested-interactive" on a11y checks. Has that been resolved? If so, there are a couple other Cypress specs that include this exception that will also need to be updated eventually.

client/test/cypress/integration/submissiondetails.spec.js
```diff
@@ -69,20 +53,13 @@ describe("Submissions Details", () => {
...
       "Application Administrator"
     )
     cy.injectAxe()
     cy.dataCy("submission_details_notify")
-    cy.checkA11y(null, {
-      rules: {
-        "nested-interactive": { enabled: false },
-      },
-    }, a11yLogViolations)
+    cy.checkA11y(null, null, a11yLogViolations)
```
gmeben, line 62: The removal of the exceptions to ignore the "nested-interactive" rule was the only questionable thing I noticed. Has this been resolved? If so, there are a couple other Cypress specs that include this rule exception that will also need to be updated eventually; this doesn't have to be done in this PR.

wreality: Removed the rest of the nested interactive options, but there are still underlying a11y issues with QSelect.

PR #1088, opened 2022-05-12
feat(api): add created_by and updated_by fields to submission

### LINKED ISSUE #1074: Add creaated_by field to submissions
### Problem/Why
There is currently no stored record of which user actually created a submission.

### Solution/How
Using the technique outlined [here](https://dev.to/hasanmn/automatically-update-createdby-and-updatedby-in-laravel-using-bootable-traits-28g9). We can also add a updated_by field, however I’m not sure that has a defined use case just yet. But, having it there wouldn’t hurt and is probably easy enough given the work involved is mostly overlapping. 

- [ ] Implement the bootable trait for auto updating user fields
- [ ] Add the created_by and updated_by fields to the submission type in graphql 
- [ ] Add unit tests on submission model to confirm fields are appropriately updated automatically (see: Carbon::setTestNow()). 
- [ ] Add feature tests to confirm graphql makes the fields available.

### Review Checklist

- [ ] Tests pass
- [ ] Tests appropriately cover added functionality

wreality:
Closes #1074


backend/database/migrations/2022_05_12_054522_add_created_by_and_updated_by_to_submissions.php
```diff
@@ -0,0 +1,53 @@
...
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('submissions', function (Blueprint $table) {
+            $createdBy = $table->unsignedBigInteger('created_by');
+            $updatedBy = $table->unsignedBigInteger('updated_by');
+
+            //SQLite doesn't like non-nullable columns having a NULL default when altering tables
```
gmeben, line 23: Is the creation of a submission considered an update and that's why `updated_by` is considered non-nullable?

wreality: Correct. Just like updated_at.

PR #1107, opened 2022-05-19
fix(client): add default slot to error banner

wreality:


wreality:
@gmeben this should fix the cypress error that’s holding up the pile of dependabot PRs. If you could review/approve it that’d be swell.

gmeben (approved):
Remarkably trivial!

Issue #1110, opened 2022-05-20
Load comments from the API


wreality:
### Problem/Why
Comments are currently mocked in the client interface and not loaded from the UI

### Goal/What
Load the comments as part of the submission graphql query.  Additionally, there is a lot of if/else complexity built into the comment components from trying to get them to play multi-duty.  We should separate the components to make them less complex and easier to maintain modify. (I know, I was a proponent of trying to combine them initially... 🤷‍♂️

### To Do / Acceptance Criteria

- [ ] Update the getSubmission query to fetch comment data
- [ ] Update components to display data returned from API
- [ ] Create inline comment component
- [ ] Create inline comment reply component
- [ ] JEST tests for added and modified components as appropriate

### Review Checklist

- [ ] Test coverage is appropriate for changes to code.

wreality:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/CCR-Development-5ff39ba363391e0019713c87/issues/MESH-Research/CCR/1110?planning-poker) with ZenHub @gmeben @sattler9 @Sruthin86

PR #1149, opened 2022-06-02
feat(client): fetch submission comments from API

wreality:
See #1110 

Still WIP, needs tests written.


wreality:
@gmeben I think we're good here now.  Review at you convenience

PR #1179, opened 2022-06-14
feat(client): display inline comment highlights on submission review page

### LINKED ISSUE #1163: Display inline comment highlights in content
### Problem/Why
The contextual location of inline comments are not represented in the content

### Goal/What
Show highlights indicating the location of inline comments in the content display on Submission Review

### Solution/How
Use #1108 as a basis to display inline comments. (Note, don't just apply it... it's well behind the git history at this point and will almost certainly explode all over the place)

- [x] Implement prototype work from #1108
~~- [ ] Add jest tests to verify functionality~~
- [x] Add cypress tests as needed as well.

### Review Checklist

- [ ] Tests added are appropriate and passing
- [ ] Comment highlight locations are indicated in the document.

### Accessibility Checklist

- [ ] Axe core passes 

NOTE: re: accessibility: Its almost a guarantee that there are going to be some UA challenges as we grow beyond "standard" interface elements and interactions.  Additional tickets to address shortcomings are acceptable remediations to accessibility concerns that would dramatically increase scope.

wreality:
Closes #1163

test(api): seed inline comments with constant data
feat(client): load content for submission review page from database

### DIFF
--- a/backend/database/seeders/DatabaseSeeder.php
+++ b/backend/database/seeders/DatabaseSeeder.php
@@ -77,9 +77,9 @@ class DatabaseSeeder extends Seeder
         $submission_file_seeder->run();
 
         $inlineCommentSeeder = new InlineCommentSeeder();
-        $inlineCommentSeeder->run(100, 1);
-        $inlineCommentSeeder->run(100, 0);
-        $inlineCommentSeeder->run(100, 10);
+        $inlineCommentSeeder->run(100, ['replies' => 1, 'highlight' => [30, 80]]);
+        $inlineCommentSeeder->run(100, ['highlight' => [430, 445]]);
+        $inlineCommentSeeder->run(100, ['replies' => 10, 'highlight' => [630, 720]]);
 
         $overallCommentSeeder = new OverallCommentSeeder();
         $overallCommentSeeder->run(100);
--- a/client/test/cypress/integration/submissionreview.spec.js
+++ b/client/test/cypress/integration/submissionreview.spec.js
@@ -61,7 +61,7 @@ describe("Submissions Review", () => {
     })
   })
 
-  it("should allow a reviewer to submit overall comments", () => {
+it("should allow a reviewer to submit overall comments", () => {
     cy.task("resetDb")
     cy.login({ email: "reviewer@ccrproject.dev" })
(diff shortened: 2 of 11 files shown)


backend/database/seeders/DatabaseSeeder.php
```diff
@@ -77,9 +77,9 @@ public function run()
         $submission_file_seeder->run();
 
         $inlineCommentSeeder = new InlineCommentSeeder();
-        $inlineCommentSeeder->run(100, 1);
-        $inlineCommentSeeder->run(100, 0);
-        $inlineCommentSeeder->run(100, 10);
+        $inlineCommentSeeder->run(100, ['replies' => 1, 'highlight' => [30, 80]]);
```
gmeben, line 80: Seeing a mixture of integer and array in the parameters makes me think this can be refactored. I tend to think specifying the submission ID is largely useless and should be removed as a parameter. If necessary, then it should be included as part of the array. What do you think?

client/test/cypress/integration/submissionreview.spec.js
```diff
@@ -61,7 +61,7 @@ describe("Submissions Review", () => {
     })
   })
 
-  it("should allow a reviewer to submit overall comments", () => {
+it("should allow a reviewer to submit overall comments", () => {
```
gmeben, line 64:
```suggestion
  it("should allow a reviewer to submit overall comments", () => {
```

gmeben (approved):

backend/database/seeders/DatabaseSeeder.php
wreality: The issue I see is there isn't a default for submissionId.  It has to be provided. Where as what's in the array has sensible defaults and can be optional.  Does that sound reasonable ?

gmeben: But there's only one seeded submission. This parameter doesn't provide any real benefit currently. For example, `SubmissionFileSeeder` just statically declares the ID. 

If you still believe the submission ID should be a parameter, it's possible to make the ID a required key in the array while having the other keys optional. 

I'm not saying you have to rewrite this. It's just a suggestion.

PR #1184, opened 2022-06-15
feat(api): Add Validation for Inline and Overall Comments

### LINKED ISSUE #1162: Add validation to reply_to_id and parentId
### Problem/Why
At the graphQL layer, a comment reply can have a reply_to_id that is not in the chain of replies it could reference (or indeed does not need to actually reference an actual existing comment it).  For parent_id, the referenced id doesn't need to belong to an existing comment or a comment existing on the submission.

### Goal/What
The mutation should fail if invalid values are provided for reply_to_id or parent_id.

### Solution/How
Add validation rules to the mutations.  Also consider adding rules attached to the laravel model itself for ensuring that parent_id is a valid value rather than relying on the Lighthouse layer to do it alone.
Rules:
- if parent_id is null, reply_to_id must be null as well
- if parent_id is NOT null:
  - the referenced comment must exist
  - the referenced comment submission_id  must equal the submission_id of the new comment,
  - the referenced comment's parent_id field must be null
- if reply_to_id is NOT null:
  - the referenced comment must exist
  - the referenced comment submission_id must equal the submission_id of the new comment
  - the referenced comment must be in the set [parent_id, ..child comments of parent_id]

- [x] Update updateSubmission.inline_comments and updateSubmission.overall_comments with validation rules.
- [x] Add tests to ensure invalid data fails.

### Review Checklist

- [ ]  New tests created appropriately cover the scope of work.

gmeben:
This pull request:

* Adds validation for `parent_id` and `reply_to_id` fields used on the creation of inline and overall comments
* Adds testing to ensure validation works as expected:
    * Inline and overall comment replies 
        * Must reply to comments that are either the parent comment or belong to the same parent comment
        * Must have parent comments that belong to the same submission
        * Must have both `parent_id` and `reply_to_id` supplied with legitimate ID references
    * Inline and overall comments 
        * Must not have either a `parent_id` and `reply_to_id` supplied
* Fixes a typo in `ArgPolicyDirective.php`

Closes #1162


wreality (approved):
Nice work!

PR #1187, opened 2022-06-17
feat(client): implement inline comment creation functionality

wreality:
This PR implements the basics of highlighting text and creating a new inline comment.

There are still some interactions that I'm not completely fond of (for ex., clicking the new comment button doesn't scroll to the new comment editor in the sidebar).

@gmeben If there's anything in this that you want to tackle right away, let do it.  Otherwise lets get it a review to get it in the codebase and start making issues to improve it.


wreality:
I'll look into fixing / refactoring failing tests in the morning.

client/src/components/forms/CommentEditor.vue
```diff
@@ -112,14 +114,26 @@ const props = defineProps({
...
+    required: false,
+    default: null,
+  },
 })
 
+const defaultContent = computed(() => {
+  if (!props.parent?.id) return ""
+  if (props.parent.id === props.replyTo?.id) return ""
+  //TODO: Make this more robust to handle multi paragraphs, etc
+  return `<blockquote>${props.replyTo.content}</blockquote>`
```
gmeben, line 128:
```suggestion
  return `<blockquote>${props.replyTo.content}</blockquote><p></p>`
```
Minimizing the chance that the user accidentally modifies the content of the blockquote and preventing them from having to hit the return key twice is preferable in my opinion.

gmeben (approved):
Here are the things I noted that need to be addressed: 

* ~~Make cancel buttons on inline comment drafting forms functional~~
* Fix scrolling on reply references from replies
* Fix replying to overall comment replies
* Make the "Scroll to Top" button functional

client/src/components/forms/CommentEditor.vue
wreality: Nice.  I like it.

PR #1202, opened 2022-06-17
feat(api): Add fields for logged in users role to publications and submissions

### LINKED ISSUE #1068: Add API fields containing current user’s role to submission, publication types
### Problem/Why
Frontend code has no simple mechanism to determine the currently logged in user’s role on a publication or submission. 

### Goal/What
Enable frontend code to be written to show/hide certain UI elements depending on a user’s role on that submission or publication. 

### Solution 

Two fields should be added to publication and submission types:
- role: the user’s specifically assigned roles on the object 
- effective_role: the user’s effective role from a permissions stand point taking into consideration their role on parent elements  

Fields should return either graphql Enums or string roles (‘submitter’, ‘reviewer’, etc)

Effective role example: if a user is a publication admin or editor their _effective_ role on a publication’s submission would be review_coordinator since they can perform all actions a review coordinator can.  

If the user is not logged in or has no role on the entity field should return null. 

- [x] Implement publicationRole method on user model. 
- [x] implement. effectivePublicationRole method on user model. 
- [x] Implement submissionRole method on user model. 
- [x] Implement effectiveSubmissionRole method on user model
- [x] Add virtual fields on publication model to fetch current user role and effective_role using Auth::user()
- [x] Add virtual fields on submission model to fetch current user role and effective_role using Auth::user()
- [x] Add fields to submission and publication graphql types
- [x] Extend submission tests to ensure entities return correct values for each field 
- [x] Extend publication tests to ensure entities return correct values for each field. 

### Review Checklist

- [ ] Tests pass
- [ ] API returns new fields when queried

wreality:
closes #1068


wreality:
Just need to add tests.  Should be simple enough to complete early next week.

backend/tests/Feature/SubmissionTest.php
```diff
@@ -549,6 +549,73 @@ public function testReviewersCannotAssignRoles($role)
...
+            ->assertJsonPath('data.submission.effective_role', 'review_coordinator');
+    }
+
+    public function testPublicationAdminGetsEffectiveRole()
+    {
+        /** @var User $user */
+        $user = User::factory()->create();
+        $this->actingAs($user);
+        $submission = Submission::factory()->create();
+        $submission->publication->editors()->attach($user);
```
gmeben, line 604: This user is assigned as an editor, not a publication administrator. I recommend one of the following:

1. Change the name of this test and add a new test case for publication admins
2. Change the role assignment in this test and add a new test case for editors
3. Add a publication roles provider and use that on this test like `allSubmissionRoles` is used elsewhere

gmeben (changes requested):
Good addition but I think testing needs some attention.

backend/app/Models/Submission.php
```diff
@@ -198,4 +198,44 @@ public function getStatusNameAttribute()
...
+        $user = auth()->user();
+
+        if (!$user) {
+            return null;
+        }
+
+        $publicationRole = $this->publication->getEffectiveRole();
+
+        if ($publicationRole !== null) {
+            return (int)Role::REVIEW_COORDINATOR_ROLE_ID;
```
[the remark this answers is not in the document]
wreality: Nope, review coordinator is their effective role on the submission.

backend/tests/Feature/SubmissionTest.php
wreality: This legit doesn't matter, but I'll update it.

gmeben (approved):

PR #1232, opened 2022-06-24
feat(client): scroll to top of inline comments when scroll to top button is clicked

### LINKED ISSUE #1200: Make the "Scroll to Top" Button Functional
### Problem/Why
The "Scroll to Top" button at the bottom of the inline comments drawer is nonfunctional.

### Goal/What
Make the "Scroll to Top" button functional so that the top of the inline comments drawer is scrolled to when when this button is clicked upon.

### Solution/How
Add a click handler for scrolling. Add the handler to the button. Write testing to ensure functionality works as expected. 

- [x] Add functionality to button
- [x] Add necessary testing

### Review Checklist

- [ ] The top of the inline comments drawer is scrolled to when the button is clicked

### Accessibility Checklist

- [ ] The button should be able to be accessed with a keyboard
- [ ] The button should be announce-able for screen readers

gmeben:
This pull request makes the "Scroll to Top" button that appears at the bottom of the inline comments drawer functional so that, when clicked upon, the drawer scrolls the user to its top in 250 milliseconds.

Closes #1200


client/src/components/molecules/InlineComments.vue
```diff
@@ -15,7 +15,13 @@
...
       />
       <div class="row justify-center q-pa-md q-pb-xl">
-        <q-btn color="dark" icon="arrow_upward">Scroll to Top</q-btn>
+        <q-btn
+          ref="scroll_to_top_button"
+          color="dark"
+          icon="arrow_upward"
+          @click="scrollToTop"
+          >Scroll to Top</q-btn
+        >
```
wreality, line 24:
```suggestion
        <q-btn
          ref="scroll_to_top_button"
          color="dark"
          icon="arrow_upward"
          @click="scrollToTop"
        >
          Scroll to Top
        </q-btn>
```

client/src/components/molecules/InlineComments.vue
```diff
@@ -42,6 +53,12 @@ const inline_comments = computed(() => {
   })
 })
 
+function scrollToTop() {
+  const target = getScrollTarget(inline_comments_section.value)
+  setVerticalScrollPosition(target, 0, 250)
+  emit("scrollInlineCommentsTop")
```
wreality, line 59:
```suggestion
```
It appears the only reason you're emitting an event is to test for it ?

client/src/components/molecules/InlineComments.vue
```diff
@@ -26,12 +32,17 @@ import { ref, watch, inject, computed, nextTick } from "vue"
 import NewInlineComment from "../NewInlineCommentComponent.vue"
 import InlineComment from "src/components/atoms/InlineComment.vue"
 import { scroll } from "quasar"
+
+const emit = defineEmits(["scrollInlineCommentsTop"])
+
```
wreality, line 37:
```suggestion
```

client/src/components/molecules/InlineComments.jest.spec.js
```diff
@@ -275,4 +275,11 @@ describe("InlineComments", () => {
     await items.at(2).find('[data-cy="collapseRepliesButton"]').trigger("click")
     expect(findReplies(items.at(2))).toHaveLength(10)
   })
+
+  test("scroll to top button triggers scrolling functionality when clicked upon", async () => {
+    const { wrapper } = wrapperFactory()
+    const button = wrapper.findComponent({ ref: "scroll_to_top_button" })
+    await button.trigger("click")
+    expect(wrapper.emitted("scrollInlineCommentsTop")).toHaveLength(1)
+  })
```
wreality, line 284:
```suggestion
```

I think the better way to test this would be to mock out the `scroll` module and `setVerticalScrollPosition` and then assert that they've been called with the appropriate element as their target.  That said, I don't think that effort is worth it in this case.  Testing an event emitted only for testing however creates a false sense of the functionality having been tested.  

If a parent component were listening for this event and dependent on it to perform some other action, then for sure, this would be a worthwhile test.

Given how much is reliant on a browser here, a cypress test may make more sense (visit page, get button, click, assert that the first comment's text is visible again) (a quick look at the docs suggests that should(be.visible) does take into account if the element is actually in the current scroll window).  Again though, I don't think going out of our way to write that test is worth the effort.

wreality (approved):
Approved assuming the changes are made.

PR #1244, opened 2022-06-27
feat(client): change application sidebar into header

wreality:
This PR removes the application level drawer/sidebar and moves its navigation elements into a header nav.


wreality:
@gmeben expecting a bunch of cypress tests to fail, feel free to ignore the PR if they do and I'll let you know when I clean them up.

gmeben (approved):

PR #1335, opened 2022-08-03
ci(client): add screenshots of translation strings to cypress

wreality:
Adds screenshots of each page in the application with both translated and bare keys displayed to aid in copyediting efforts.

TODO:
- [x] In the copy editing cypress test, add the 3rd parameter to routes that need to wait on the existence of an element before taking the screenshot.
- [x] Update docs with a section about this feature and how to access the screenshots (can also make this a separate issue instead)


wreality:
Added the artifact upload to the workflow file so for example https://github.com/MESH-Research/CCR/actions/runs/2806044269 now list screenshots in the artifacts for download.

@gmeben I also added the TODO items to the description, adjust as you see fit.

gmeben (approved):

PR #1337, opened 2022-08-04
feat(client): Add Fitness for Review Interactive Elements to User Interface

### LINKED ISSUE #754: Determine Fitness for Review
### Problem/Why
According to the ["Submission workflow" Miro board](https://miro.com/app/board/o9J_l654lxo=/), a submission is meant to be evaluated for its fitness for review before being able to be reviewed by reviewers. This is currently not possible in CCR.

### Goal/What
Immediately after a submission is created, implement a way for editors to designate if a submission is fit for review. These should be the allowed choices: 

1. Accept
2. Request Resubmission
3. Reject

Reviewers should be prevented from reviewing the submission until this fitness for review designation is made by an editor. 

### Solution/How
Update Vue templates according to the wireframes in issue #1271

Add mutation functionality to interact with the API that changes the status of the submission based on which of the following options is selected:

* Accept > Awaiting Review
* Request Resubmission > Awaiting Resubmission
* Reject > Rejected

Update the Vue template so that the button that allows a user to review the submission is only interactive to reviewers when the submission is accepted for review. Ensure this permission checking is also on the route to prevent direct access via URL.

- [x] Update the template files (interactive elements, notifications)
- [x] Add mutation functionality for changing the submission status on the GraphQL API from the client
- [ ] ~~Add permissions for viewing a submission out of review~~
- [x] Add permission checking rules to prevent unauthorized access from reviewers
- [x] Add Cypress testing to ensure interactions work as expected

### Review Checklist

- [ ] An editor can accept a submission for review
- [ ] An editor can reject a submission for review
- [ ] An editor can request a resubmission on a submission
- [ ] Reviewers can access the submission review page only when the submission is accepted for review

### Accessibility Checklist

- [ ] The controls for determining a submission's fitness for review are announceable via screen reader
- [ ] The text for the controls exists in the i18n.js file
- [ ] The language for the controls uses [plain language](https://sheribyrnehaber.com/plain-language-essential-for-content-accessibility/)

gmeben:
This pull request:

* Adds a new dialog for changing the status of a submission with the following actions:
    * Accept for Review -> Awaiting Review
    * Request Resubmission -> Resubmission Requested
    * Reject -> Rejected
* Adds a nested menu on the Submissions page that utilizes the new dialog
* Adds buttons to the toolbar of the Submission Review page that utilizes the new dialog
* Denies reviewers access to the Submission Review page for submissions that are in a non-reviewable state
* Renames "View Submission" to the more appropriate "Review Submission"

Closes #754


backend/database/seeders/SubmissionSeeder.php
```diff
@@ -31,10 +31,22 @@ public function run()
      *
      * @param int $id
      * @param string $title
-     * @return void
+     * @param array $data
+     * @return \Database\Seeders\App\Models\Submission
      */
-    protected function createSubmission($id, $title)
+    protected function createSubmission(int $id, string $title, array $data = [])
```
wreality, line 37: I changed this up a little bit to accept data as an optional third parameter rather than requiring an extra save.

client/src/apollo/apollo-router-guards.js
```diff
@@ -53,6 +53,72 @@ export async function beforeEachRequiresSubmissionAccess(
   }
 }
 
+export async function beforeEachRequiresReviewAccess(
```
wreality, line 56: I think there's rework we need to do here with how we're checking for access by role + if the submission is open.  Some of this should perhaps be relocated to a mounted hook on the component, but we need to figure out where the separation of concerns should lie between router hooks and the mounted hook on the component.

(Just a note for future, it works now, I think its just going to get messier as the application progresses if we don't address it though.)

wreality (approved):
Just a couple of notes and a couple of small refactors.  I added a comment as well about refactoring/rethinking when we use router hooks / cleaning some of that up.  We don't need to address that now though.

Excellent work @Sruthin86 @vilagshe & @gmeben

PR #1366, opened 2022-08-19
build(deps): Update cypress to version 10

wreality:
* Updates cypress to version 10
* Squashes migrations into a schema file
* Adds seeder for roles/permissions from removed migrations.


wreality:
Everything appears to be working well and there was only a modest increase in run time, so this should be good for review / merge.

backend/database/seeders/RoleSeeder.php
```diff
@@ -0,0 +1,67 @@
...
+    {
+        $roles = Role::getArrayOfAllRoleNames();
+        foreach ($roles as $key => $role) {
+                Role::factory()->create([
+                    'id' => $key + 1,
+                    'name' => $role,
+                ]);
+        }
+
+        Permission::create(['name' => Permission::UPDATE_USERS])
```
gmeben, line 27: Since this class is seeding both roles and permissions, I think it should either be:

1. broken into two different classes: `RoleSeeder` and `PermissionSeeder` with the permission seeding code extracted
2. renamed to `RoleAndPermissionSeeder.php`

gmeben (approved):

backend/database/seeders/RoleSeeder.php
wreality: Totally see where you're coming from, but I don't think its a big deal here.  I'll keep a rename in mind as this seeder evolves though.

PR #1377, opened 2022-08-25
feat: Add validation for style criteria submitted with inline comments

### LINKED ISSUE #1158: Verify validity of style criteria submitted with inline comments
### Problem/Why
Any data can be submitted for icon and name fields for new/modified inline comments.

### Goal/What
The inline comment style criteria field should only accept style criteria that are present on the publication.

### Solution/How
Update style_criteria input on InlineComments to accept an optional array of IDs.  The backend should then use an [ArgTransformer](https://lighthouse-php.com/5/custom-directives/argument-directives.html#argtransformerdirective) directive to tranform the data into a JSON-serializable array.  The data stored in the database should not change as a result of this modification.

New style criteria: 
```
style_criteria: [ID!]
```

- [x] Update schema and create directive
- [x] Update existing tests to supply correct graphql data.

### Review Checklist

- [ ] Tests pass

gmeben:
This pull request changes the requirement for supplied style criteria submitted with inline comments to be a list of IDs that map to style criteria that exist in the database instead of an object with `name` and `icon` properties of arbitrary values. The data stored in the database does not change as a result of this modification.

Closes #1158


wreality (approved):
This looks like exactly what I was hoping we'd be able to implement.  Nice work!

PR #1397, opened 2022-08-31
fix(client): Update client for new expected input format

gmeben:
This updates the client to use the new expected input format for style criteria (an array of IDs) that are submitted on submissions. It also removes a `console.log()` statement.


wreality (approved):
Fix looks good, thanks!

We should make a note/issue that:

a) The client doesn't show an error if a graphql call fails (perhaps a catchall network error/etc/etc that just popups up a "An error occurred communicating with the application" or something like that)
b) We need testing for inline comment writing that would catch this regression in the future.

PR #1402, opened 2022-09-02
fix(client): update deep selectors in comment reply quotes (deprecation)

wreality:
Fixes warning generated during a client build: `[@vue/compiler-sfc] ::v-deep usage as a combinator has been deprecated. Use :deep(<inner-selector>) instead.`


gmeben (approved):
It bothers me that simply `blockquote` doesn't work as a selector in this case but I'm not able to suggest a better alternative at this time.

PR #1446, opened 2022-09-15
build: Fix testing container for M1 macs

gmeben:
This fixes an issue where attempting to run Cypress (integration testing) on an M1 Mac inside the cypress lando container would fail. It upgrades the version of Node that Cypress uses from 14 to 16. 

This fix was made possible by the development mentioned in this comment of this issue: https://github.com/cypress-io/cypress-docker-images/issues/431#issuecomment-1184977847


wreality (approved):
I just updated the description a bit.  Otherwise GTG

Issue #1246, opened 2022-06-30
Convert File Upload for Submissions


gmeben:
### Problem/Why
Users are not able to upload a word document to the application and review its contents as a submission because uploaded files are not convert.

### Goal/What
Make the content of word document file uploads converted to HTML so they can be reviewable as submissions on the Submission Review page.

### Solution/How
After a word document is uploaded, use `pandoc` to extract the content of the file (if supported) and convert this content into HTML to be used as the submission content data for its newly created submission. This should correctly populate the `submission_file_id` database field in the `submission_contents` table.

This conversion should utilize the research findings from issue #517 in terms of what file types are supported by Pandoc for intake. 

**Supported Word Document Input Formats**

See the [Editorial Meeting Rolling Notes from July 6th, 2021](https://docs.google.com/document/d/1rd81MjhbUH1B0czjI0BYxR0U04AOidJfsisbulJVj2Q/edit#heading=h.nrofhw76j0be) for details on preferred input formats. 

* docbook
* docx (preferred)
* epub
* html
* latex (preferred)
* md
* odt (preferred)
* rtf

Because conversion can be a time-consuming process, it should be expected to be an operation that runs asynchronously to the creation of the submission.

### To Do / Acceptance Criteria

- [x] Add code that checks the filetype of an uploaded file to compare against the supported input file types
- [x] Use pandoc to convert the file's contents into HTML
- [x] Create a record in the `submission_contents` table with the newly converted HTML and with the correct foreign key IDs populated

### Review Checklist

- [ ] ~~The content of uploaded word document files is converted into HTML~~
- [ ] ~~The content of uploaded word document files can be accessed on the corresponding submission's Submission Review page~~

gmeben:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/CCR-Development-5ff39ba363391e0019713c87/issues/MESH-Research/CCR/1246?planning-poker) with ZenHub @sattler9 @wreality

PR #1517, opened 2022-10-03
feat(api): Convert uploaded files to html content

wreality:
This PR processes newly uploaded files using Pandoc, creates a new content entity and associates the content entity with the submission as its current content.  On error, the file entity is marked with as having an error and the error message is added to the entity for future debug/consideration.  If a file type that is not convertable by pandoc is uploaded, an UnknownInputFormat exception will be stored for the file entity and the content will not be updated.

The PR doesn't change anything related to how files are processed on the front end or the file upload process itself.  We'll need to add validation on the file types uploaded in order to prevent files from even being processed by pandoc, but with upcoming changes to the UI it seemed better to consider those then.

#1246


backend/tests/Feature/SubmissionContentTest.php
```diff
@@ -69,6 +69,7 @@ public function testSubmissionContentCanBeAccessedFromASubmissionFile()
                 'submission_id' => $submission->id,
             ]);
             SubmissionFile::factory()->create([
+                'file_upload' => '/tmp/testfile.txt',
```
wreality, line 72: Specify an uploaded filename since this test doesn't rely on an actual file.

gmeben (approved):

PR #1556, opened 2022-10-20
fix(client): Fix bug preventing user account pages from loading

wreality:
This PR fixes should fix the majority of existing cypress issues.  I think there still may be some issues in individual local devs that we still needs to track down, but this did pass everything locally for me.


client/src/layouts/Account/AccountLayout.vue
```diff
@@ -2,7 +2,7 @@
   <div>
     <div class="row justify-center items-start content-start q-pa-md">
       <q-card class="col-sm-3 col-xs-12 no-shadow no-border-radius">
-        <div class="row">
+        <div v-if="currentUser" class="row">
```
gmeben, line 5: This looks right to me.

gmeben (approved):

PR #1623, opened 2022-11-21
feat: Limit comment modification to authors only

### LINKED ISSUE #1542: Limit comment modification to comment authors only
### Problem/Why
Currently inline and overall comments can be modified by any user. 

### Goal/What
Enable comments to be modified only by the user that authored the comment. Other users should not be able to modify comment text for comments they did not author. 

### Solution/How
Add checks when a user clicks on "Modify Comment" to ensure that user is the comment author before allowing the comment modification mutation to execute. Disable the modification option for non-authors and provide a tooltip.

* [x] Add checks on the comment components
* [x] Add checks in the backend to ensure only the author of a comment can modify that comment
* [x] Add Cypress tests

### Review Checklist
* [x] Comments can be modified by their author
* [x] Comments cannot be modified by other users
* [x] The "modify comment" menu option is not available for non-authors

### Accessibility Checklist
* [x] Cypress testing passes

vilagshe:
Closes #1542 

- Limits comment modification functionality to comment creators/authors only
  - Adds user checks in front end
  - Adds user checks in back end
- Adds cypress tests


client/test/cypress/integration/submissionreview.cy.js
```diff
@@ -297,39 +312,67 @@ describe("Submissions Review", () => {
...
   })
 
-  it("should allow inline comments to be modified", () => {
-    cy.task("resetDb")
-    cy.login({ email: "applicationadministrator@ccrproject.dev" })
-    cy.visit("submission/review/100")
+  // it("should allow inline comments to be modified", () => {
+  //   cy.task("resetDb")
+  //   cy.login({ email: "applicationadministrator@ccrproject.dev" })
+  //   cy.visit("submission/review/100")
```
gmeben, line 318: This function should be un-commented.

client/src/components/atoms/CommentActions.vue
```diff
@@ -35,3 +36,26 @@
...
+const { currentUser } = useCurrentUser()
+
+const comment = inject("comment")
+
+defineEmits(["quoteReplyTo", "modifyComment"])
+
+const checkCommentCreatedBy = computed(() => {
+  const userToCheck = currentUser.value.id
+  const commentCreatedBy = comment.created_by.id
+  console.log(userToCheck, commentCreatedBy)
```
gmeben, line 53:
```suggestion
```

gmeben (changes requested):
I'm not able to modify comments successfully on my local environment. I keep encountering a server-side error about `commentFields` being missing. There's also some leftover output being logged to the console. 
Numbers and an error logged in the console that reads, "Error: Unknown fragment 'commentFields'. Unknown fragment 'commentFields'."

client/test/cypress/integration/submissionreview.cy.js
vilagshe: In order for this test to pass, it would require creation of a new inline comment via cypress, which we discussed the difficulty of doing. Inline comment creation is also not being tested in this spec, so we discussed commenting it out.

vilagshe:
> I'm not able to modify comments successfully on my local environment. I keep encountering a server-side error about `commentFields` being missing. There's also some leftover output being logged to the console. Numbers and an error logged in the console that reads, "Error: Unknown fragment 'commentFields'. Unknown fragment 'commentFields'."

I'm getting this same warning only for overall comment modification. Looking into it, but noticing that cypress tests are passing for testing overall comment modification.

client/test/cypress/integration/submissionreview.cy.js
gmeben: My takeaway from our discussion is that seeded inline comments would fulfill the needs of this test and that creation isn't necessary. The name of this test is `should allow inline comments to be modified` which is undoubtedly pertinent and satisfies the need to ensure that 1 of the 4 comment types that can be modified can be done so successfully.

vilagshe: The problem with seeded comments is that the comment authors are currently generated randomly, so there is no way to always point to a comment that can be modified. We discussed the difficulty in creating a new inline comments using cypress to get around this, and ultimately decided that because Inline comment creation is not being tested either, commenting it out was an option.

gmeben: It's possible to update the seeding so that you can guarantee you have a comment to work with. If you need help, I'd be willing to assist and pair program.

vilagshe: I thought we decided to not modify the seeder file, but I can make those changes again so that the inline comments that are seeded have defined users.

vilagshe:
> 

I do not know why this error is only happening for Overall Comment modification. As far as I can tell, it's set up the same as the other comment types, which are able to be modified fine. I'm open to working together to try to figure this out.

client/test/cypress/integration/submissionreview.cy.js
gmeben: I apologize for leaving the incorrect impression. Yeah, updating the seeder would be one of the simpler approaches I think.

gmeben:
By the way, I encountered this error while attempting to modify each of the four different comment types. Feel free to direct message me with a good time to meet.

gmeben (approved):
I'm able to modify all four comment types locally now. Good work.

PR #1773, opened 2023-03-07
feat: Add Basic Elements to the Dashboard Page

### LINKED ISSUE #1732: Add Basic Elements to My Dashboard Page
### Problem/Why
When any user logs in to the application, they are redirected to a page titled "My Dashboard" that is completely blank. This is the experience of submitters, reviewers, review coordinators, editors, and administrators. Therefore, this page is not useful and can give the impression that the application is incomplete.

### Goal/What
Provide basic useful elements on the My Dashboard page that will serve users of all roles, including submitters, reviewers, review coordinators, editors, and administrators. 

These elements can include:

* Call to action to visit [the Guide section of the documentation site](https://ccr.meshresearch.dev/guide/) for new users seeking to learn how to use the application 
* A list of submissions associated with the user logged-in. Those associations can be:
    - as a submitter
    - as an assigned reviewer
    - as an assigned review coordinator
    - as an editor of the publication the submission is associated with
    - as a publication administrator of the publication the submission is associated with

### Solution/How
Add backend functionality as needed to retrieve pertinent information for the client. Update the My Dashboard Vue template with the additional elements specified above and add GraphQL queries as necessary to populate those elements with real data.

- [ ] Add functionality to the backend
- [ ] Update the My Dashboard Vue template

### Review Checklist

- [ ] The My Dashboard page contains elements that any user will find useful
- [ ] While the My Dashboard page may not contain all information ultimately desired, the My Dashboard page does not give the impression that the application is incomplete

### Accessibility Checklist

- [ ] Interactive elements can be accessed solely by keyboard
- [ ] Interactive elements can be read by a screen reader
- [ ] Any added copy uses [plain language](https://sheribyrnehaber.com/plain-language-essential-for-content-accessibility/)
- [ ] No new accessibility violations are introduced

gmeben:
This pull request:

* Adds the following to the Dashboard:
    - Welcome section with links to the account and profile pages
    - Link to the [user guide](https://pilcrow.meshresearch.dev/guide/)
    - Conditionally visible tabular lists of the authenticated user's reviews, coordinations, and submissions
* Makes the `AppHeader` better fit smaller viewports by 
    - toggling the visibility of the application title
    - adjusting the padding size of navigational items
* Improves the `SubmissionTable` component
    * Adds a `variation` prop to enable more customization
    * Adjusts the border style and ensures this is consistent between light and dark mode
* Fixes the appearance of links and text that use the primary and accent colors for dark mode

Closes #1732


vilagshe (commented):
the dashboard looks good! I did notice a few things.

- The table columns change width when a filter is applied.
dashboard-before-filter
dashboard-after-filter

- The table is not optimized for mobile, and requires a horizontal scroll for smaller screen sizes.
- Also on mobile, the filters dropdown looks like it moves to the middle/left side of the page while it is on the right side on desktop. I feel like this should remain right-aligned with the table once optimized for mobile.

- Only one status can be filtered for at a time. Do we want to be able to select multiple?

gmeben:
@vilagshe 

1. I added a relative width to the Submission Title column which should make column widths more consistent responsively. 
2. I'm now conditionally applying [Quasar's grid style table](https://quasar.dev/vue-components/table#grid-style) for small viewports. 
3. Not only do I find this left alignment to not be problematic, I think it's superior to the proposed right-aligned alternative which appears janky. Am I reading your comment correctly or did you mean something different? 

screenshots comparing the left aligned and right aligned versions


4. That's a good idea. I will see what I can do to make this [multi-selectable](https://quasar.dev/vue-components/select#example--generating-multiple-values).

vilagshe:
1. This fixed the column width issue!
2. I like the look of the table in the screenshot, I think that will work better for mobile and smaller screen sizes
3. The dropdown is on the right of the page on desktop, so I was expecting it to be in the same place on mobile, and the switch in sides surprised me when I changed screen sizes, which is why I indicated that maybe it should stay on the right. With the new table design for mobile, it does look better on the left, so perhaps it should now also be on the left in desktop. Whichever way, I think it needs to stay on the same side of the page between desktop/mobile.

gmeben:
@vilagshe 

I've finished the `item` slot so that the grid style layout works for small viewports and fashioned it to match [your original design](https://xd.adobe.com/view/dbe1b864-2527-41d2-a2d9-f854bd22fe66-8eb2/screen/bd8ac0a1-a2b6-4404-8229-49edafc090b3/) as closely as I could. I also got multi-selection implemented and adjusted the position of the filter input so that it stays on the same side of the page between desktop and mobile. 

Here's what that looks like on the Reviews page for desktop. 

screenshot of the submission table on the Reviews page

Here's what that looks like on the Dashboard page for medium sized viewports.

dashboard-medium

And here's what that looks like on the Dashboard page for tiny sized viewports.

dashboard-tiny

vilagshe:
This looks great! My only comments now are:
1. It feels a little tight, and could use more whitespace between rows
2. Having the status right aligned feels correct for screen sizes where it can fit on the same row as the publication name, but seeing it on a smaller screen feels like it should be left aligned.

vilagshe (approved):
I'm going to approve this now, since we've covered most of the feedback already

gmeben:
@vilagshe 

1. I'm surprised you feel that way. I would need to see an example of how much more whitespace should be added. I think that's something that could be addressed separately. 
2. Here's how that suggestion works:

screenshot of the Dashboard on a small viewport demonstrating inconsistent text wrapping for the record whose status is Resubmission Requested because its status name is too long

The problem is that, depending upon the viewport and the length of the status name, text wrapping can occur inconsistently and create a troubling user experience. In keeping with the spirit of "elements staying in the same spot between desktop/mobile," I anticipated that you would prefer that the status always be right aligned so I took extra time to make it work that way. If you feel like this still needs to be reworked, I would recommend we examine this more closely for the Submissions Page redesign.

PR #1795, opened 2023-03-24
chore: replace client build system with vite

wreality:
This change removes quasar/app-webpack and replaces it with quasar/app-vite. `lando rebuild` will be required for developers, and clearing out `client/node_modules` isn't a bad idea either with all of the dependency changes that come along with this.

## Remaining Todos

- [x] Two JS unit tests are still failing
- [x] Loads of console output in vitest output
- [x] One cypress test is failing
- [x] Add eslint-plugin-vitest to replace eslint-plugin-jest
- [x] Port non component jest tests to vitest (see linter output)
- [x] Delete client.old from .gitignore
- [x] Remove majestic (specific to jest)


wreality:
I noticed that about lando as well. I might try to dig into that in the future. What other issues are you experiencing ?

Also yes, the name of the test suite changed from jest to vitest which impacts the required list of tests. Once it's merged I'll fix it up.

gmeben:
@wreality Oh, those were the only issues I experienced. I was just characterizing them as "some strange behavior" to summarize my experience overall. 

If you're interested in what my Docker containers are doing specifically when starting up, I'd be willing to share my screen at some point.

wreality:
If it's doing what it should from your perspective other than the health check from lando, I'm happy.

PR #1854, opened 2023-06-15
feat: add ux testing seeder

wreality:
This PR creates a seeder for creating UX testing users/content/etc.  

Users 1-6: (Replace 1 with the number desired, passwords are all the same) 
tester1@pilcrow.dev
pilcrowRocks!@#

Submissions (Replace 1 with the number desired, each has the corresponding testing user as a reviewer): 
UX Test 1 Submission

For reviewer: 
After starting up pilcrow run
`lando artisan migrate:fresh --seed` to run the standard migrations
`lando artisan db:seed UxSeeder` to run the UX migrations

Note: this seeder will only run when manually run from the command line, a feat the staging build server will handle.

### DIFF
--- a/backend/database/seeders/UxSeeder.php
+++ b/backend/database/seeders/UxSeeder.php
@@ -1,0 +1,1168 @@
+<?php
+declare(strict_types=1);
+
+namespace Database\Seeders;
+
+use App\Models\Submission;
+use App\Models\SubmissionContent;
+use App\Models\User;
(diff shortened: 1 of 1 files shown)


gmeben (approved):
I was able to log in as a UX tester and access one of the submission versions. 

Unfortunately, the anchor links between references and their footnotes/endnotes remain nonfunctional, but I believe that's more of a problem with Pandoc.

PR #1855, opened 2023-06-15
chore: do not publish docker images on pull requests or rennovate action

wreality:
This PR adds a conditional to the publishing of container images to help prevent build errors on pull requests.


gmeben (approved):
I'm not sure how to go about testing this, but the [code usage here](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context) looks correct so I'll approve.

PR #1874, opened 2023-07-01
ci: fix duplicate action runs on rennovate pull requests

wreality:


wreality:
The concurrency part of this works, but the skipped tests make the PR status failing which doesn't work for branch protections.

image

Ref: https://github.com/orgs/community/discussions/8336

PR #1882, opened 2023-07-01
fix: downgrade dependency graphql/graphql-js to 16.6.0

wreality:
This reverts commit 921a31507ccbf43df7c2fc9ebc14fcd28ec6a6b1.

A bug exists in graphql 16.7.1 that causes an error when quasar is running in development mode.   This doesn't affect the build or the compiled assets so it missed detection during the automated testing.  

Ref: https://github.com/graphql/graphql-js/issues/3925


wreality:
Approved via workflow so that things work on Monday :-D

PR #1903, opened 2023-07-14
feat: import foot/endnotes from docx files

wreality:


gmeben (approved):
@wreality I added the Cypress testing you requested. It asserts that click/scroll functionality works as expected for footnotes and their references. 

I was getting inconsistent test results with `footnote_test.docx` because my test expects after clicking on a footnote reference link that the viewport's position changes more dramatically than it does for `footnote_test.docx`. 

Since the layout of the Submission Review page is likely to change later, I figured it's easier to make the test content less ambiguous to work with than to try to engineer something more agreeable with the layout. That's why I added `footnote_test_with_lorem_ipsum.docx`. I'm totally fine if you'd rather take a different approach and change this, but I hope my contribution at least provides some useful ideas. 

As a bonus, I also fixed a click handler mixup in `SubmissionDraft.vue`.

PR #1961, opened 2023-09-08
feat: Add backend protection for comment modification

### LINKED ISSUE #1924: Backend Protection for Comment Modification
While there are protections against comment modification on the frontend, it is still possible for a comment to be modified by an authenticated user sending post requests to the GraphQL endpoint directly for comments that don't belong to them. 

Add protection to the backend so that this cannot be done. 

**Review Checklist** 

* Necessary backend unit testing has been added 
* Necessary unit testing has been added/updated

gmeben:
This pull request ensures that users may only modify their own comments, even when modifying directly via the GraphQL endpoint

Bonus: 

* This makes some dataProviders `static` and helps avoid new warnings as [PHPUnit v10.3 now requires data provider functions to be static](https://docs.phpunit.de/en/10.3/writing-tests-for-phpunit.html#data-providers)

Closes #1924


gmeben:
The previous validators have now been refactored as policies. @wreality

wreality (approved):

PR #2022, opened 2023-11-13
build: Use more Cypress-compatible Node version

gmeben:
This prevents errors related to Node 20 being (somehow) incompatible with the application's instance of Cypress.

### DIFF
--- a/.lando.extras.yml
+++ b/.lando.extras.yml
@@ -3,9 +3,9 @@ cypress:
   template:
     services:
       cypress:
-        type: node:20
+        type: node:18
         overrides:
-          image: "cypress/base:20"
+          image: "cypress/base:18"
           environment:
             ELECTRON_EXTRA_LAUNCH_ARGS: "--force-prefers-reduced-motion"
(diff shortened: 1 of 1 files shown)


.lando.extras.yml
```diff
@@ -3,9 +3,9 @@ cypress:
   template:
     services:
       cypress:
-        type: node:20
+        type: node:18
         overrides:
-          image: "cypress/base:20"
+          image: "cypress/base:18"
```
wreality, line 8: Didn't we determine this needed to be a specific version b/c `cypress/base:18` doesn't exist?

wreality (approved):
Approving with the caveat that the version for cypress might not be right.  I don't need a second look once that's fixed though.

.lando.extras.yml
gmeben: Oh! `cypress/base:18.16.1`, right?

gmeben: Changed as suggested.

Issue #2046, opened 2024-01-08
Add a list of the types of documents accepted for submission


sattler9:
To aid the submitter, it would be kind and helpful to let them know which document types are accepted by Pilcrow prior to trying to submit a file. This will prevent frustration or the belief that Pilcrow is broken when a type that is not accepted is used in the submission process and fails.

**Describe the solution you'd like**
Add a list of the document types that Pilcrow accepts to the Submission Content page in the upload box.

<img width="1569" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBelNxQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--f7f07afa2f8620f6bb610e626683615a62f2864c/Screenshot%202024-01-08%20at%203.43.51%E2%80%AFPM.png" alt="Screenshot 2024-01-08 at 3.43.51 PM.png" />

This is associated with the feature request at https://feedback.pilcrow.dev/posts/19/upload-types

sattler9:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Pilcrow-Development-5ff39ba363391e0019713c87/issues/MESH-Research/Pilcrow/2046?planning-poker) with Zenhub @gmeben @sattler9 @vilagshe @wreality

gmeben:
Related issue: https://github.com/MESH-Research/Pilcrow/issues/517

PR #2059, opened 2024-01-19
update: add list of accepted document types

vilagshe:
This pull request:

- Adds text describing accepted document types for submission uploads

Closes: #2046

### DIFF
--- a/client/src/i18n/en-US.json
+++ b/client/src/i18n/en-US.json
@@ -892,11 +892,15 @@
       "heading": "Submission Content",
       "upload": {
         "label": "Upload",
-        "caption": "Select a supported file to be imported."
+        "caption": "Select a supported file to be imported.",
+        "file_caption": "Accepted file types: ",
+        "file_types": ".csv, .docbook, .docx, .epub, .html, .latex, .md, .odt, .rtf"
       },
       "upload_update": {
         "label": "Upload",
-        "caption": "Select a supported file to replace the existing text content."
+        "caption": "Select a supported file to replace the existing text content.",
+        "file_caption": "Accepted file types: ",
+        "file_types": ".csv, .docbook, .docx, .epub, .html, .latex, .md, .odt, .rtf"
       },
       "enter": {
         "label": "Enter Text",
--- a/client/src/pages/SubmissionContent.vue
+++ b/client/src/pages/SubmissionContent.vue
@@ -58,6 +58,7 @@
             >
               {{ $t(`submissions.content.upload${updateModifier}.caption`) }}
             </div>
+
             <template v-if="updateMethod !== ''" #action>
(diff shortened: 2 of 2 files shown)


client/src/i18n/en-US.json
```diff
@@ -892,11 +892,15 @@
...
+        "caption": "Select a supported file to be imported.",
+        "file_caption": "Accepted file types: ",
+        "file_types": ".csv, .docbook, .docx, .epub, .html, .latex, .md, .odt, .rtf"
       },
       "upload_update": {
         "label": "Upload",
-        "caption": "Select a supported file to replace the existing text content."
+        "caption": "Select a supported file to replace the existing text content.",
+        "file_caption": "Accepted file types: ",
+        "file_types": ".csv, .docbook, .docx, .epub, .html, .latex, .md, .odt, .rtf"
```
gmeben, line 903: I'm not a fan of this duplication. I can't imagine these `file_caption` and `file_types` values ever needing to be different between uploading or updating an upload like I can for the other values. 

If you want to be thorough, I think this means refactoring `upload` and `upload_update`. If you want a quick way to do this, then I recommend rewriting this so that `file_caption` and `file_types` are combined into one new node, like at `content.support` for example.

client/src/pages/SubmissionContent.vue
```diff
@@ -95,6 +96,28 @@
             </template>
           </q-banner>
           <div v-if="updateMethod == 'upload'">
+            <q-card class="q-pa-md" flat>
+              <q-list>
+                <q-item clickable>
```
gmeben, line 101: I don't think this needs to be clickable.
```suggestion
                <q-item>
```

client/src/i18n/en-US.json
```diff
@@ -892,11 +892,15 @@
       "heading": "Submission Content",
       "upload": {
         "label": "Upload",
-        "caption": "Select a supported file to be imported."
+        "caption": "Select a supported file to be imported.",
+        "file_caption": "Accepted file types: ",
+        "file_types": ".csv, .docbook, .docx, .epub, .html, .latex, .md, .odt, .rtf"
```
gmeben, line 897: I just double-checked by testing it and `.txt` files are supported as input types despite [the findings I documented here](https://github.com/MESH-Research/Pilcrow/issues/517#issuecomment-1105478040). I don't know why I documented they were output only. Maybe that was the case at the time and it changed since then? In any case, I apologize for the confusion. 
```suggestion
        "file_types": ".csv, .docbook, .docx, .epub, .html, .latex, .md, .odt, .rtf, .txt"
```

gmeben (changes requested):

gmeben (approved):

PR #2093, opened 2024-03-21
fix: Update en-US.json

gmeben:
Fix a couple typos


gmeben:
This fixes a couple typos on user-facing copy.

PR #2113, opened 2024-04-29
test: Check Accessibility on Forms

### LINKED ISSUE #193: Check Accessibility on Form Validation Messages
### Problem/Why
The current implementation of a11y accessibility checking in Cypress checks a limited set of interfaces and does not check all of the different states of validation messages on forms.

### Goal/What
Add Cypress accessibility checks for the different states of forms.

### Solution/How
Within Cypress specs that check forms (Register, Login), add `cy.checkA11y()` checks throughout to test the accessibility of different validation messages.

- [x] Identify areas on the Register spec where intermediate states appear and add checks
- [x] Identify areas on the Login spec where intermediate states appear and add checks

### Review Checklist

- [ ] All different states of each form's appearance have an accessibility check 

### Accessibility Checklist

- [ ] All Cypress axe tests pass

gmeben:
* Adds missing a11y checks in various places, including: forms, the Reviews page, the Login page, the Register page, the Submission Create page, and the Submission Details page
* Fixes an accessibility issue with QItems within QMenu
* Fixes missing headings on the Profile and Settings pages
* Bonus: Refactors code for handling dark mode into a composable

Closes #193


client/src/use/guiElements.js
```diff
@@ -1,8 +1,23 @@
...
+    () => {
+      darkModeStatus.value = $q.dark.isActive
+    },
+  )
+  function toggleDarkMode() {
+    $q.dark.toggle()
+  }
+  return { darkModeStatus, toggleDarkMode }
+}
+
```
wreality, line 20: So, pardon if I'm missing something, but why wouldn't we just use `const { dark: { toggle, isActive} = useQuasar()` in the component rather than wrapping this in our own composable?

wreality (changes requested):
Just one question in the comments, otherwise all looks good.

PR #2120, opened 2024-07-08
fix: move permission creation to a migration.

wreality:


gmeben (approved):

backend/database/seeders/DatabaseSeeder.php
```diff
@@ -1,4 +1,5 @@
 <?php
+
```
gmeben, line 2:
```suggestion
```

backend/database/seeders/RoleSeeder.php
```diff
@@ -1,4 +1,5 @@
 <?php
+
```
gmeben, line 2:
```suggestion
```
Linter doesn't like this

gmeben (approved):

PR #2122, opened 2024-07-26
feat: Add 'Mark Read' Functionality for Comments

### LINKED ISSUE #2071: Add Indicators For New Replies
If another user replies to a comment, there is no explicit indicator that the reply was newly added or that the reply has been seen before.

Add functionality to detect whether or not a reply has been viewed by a user and show some visual indication of such.  Perhaps this should be tied into a corresponding notification being read?

**Review Checklist**

* Visual indicators of a newly added replies are visible
* Necessary frontend unit testing has been added
* Necessary integration testing has been added

**Accessibility Checklist**

* No new color contrast issues are introduced
* Changed text has a font size of at least 16px
* Changed text uses plain language

gmeben:
* Adds reference buttons to all inline and overall comments to indicate unread status
* Updates tooltips on references based on comment type and read status
* Enables users to mark unread inline and overall comments as read by clicking/tapping on reference buttons
* Makes comment focus styling for "active comments" consistent with solid stroke lines

Screenshot of the green highlight appearance of an unread inline comment as it appears in light and dark mode

Closes #2071


backend/app/Models/Traits/ReadStatus.php
```diff
@@ -0,0 +1,62 @@
+<?php
```
wreality, line 1: Really like the use of a trait here.

client/src/components/atoms/CommentReference.vue
```diff
@@ -0,0 +1,83 @@
...
+    content = t(`submissions.comment.reference.mark_read`)
+  } else if (props.comment.__typename === "InlineComment") {
+    content = t(`submissions.comment.reference.go_to_highlight`)
+  }
+  return content
+}
+
+function setActive() {
+  //Null the active comment first to trigger the scroll watcher
+  //TODO: Do this in a more elegant way.
```
wreality, line 77:
```suggestion
  //TODO: Fix active comment updates needing nextTick (reactivity)
```
Trying to make sure I make the actual todo comment something understandable when using a tool like https://github.com/Gruntfuggly/todo-tree

wreality (approved):
Looks good.  Appreciate the cruft cleanup you did as part of this as well.  🚀

backend/app/Models/Traits/ReadStatus.php
gmeben: You originally developed that so I should hope you like it!

PR #2240, opened 2026-03-20
feat(dev): show local dev folder name when not default

wreality:
## Summary
- Adds a `LocalDevBanner` component that displays the Lando app root folder name as a small pill at the top of the header
- Only visible when the folder name differs from the default (`pilcrow`), helping developers distinguish between worktrees/checkouts during local development
- Passes `LANDO_APP_ROOT_BIND` env var through Quasar config
image

## Test plan
- [x] Run in Lando from the default `pilcrow` directory — banner should not appear
- [x] Run in Lando from a differently-named directory — banner should show the folder name

### DIFF
--- a/client/src/components/LocalDevBanner.vue
+++ b/client/src/components/LocalDevBanner.vue
@@ -17,9 +17,10 @@ const dirName = appRoot?.split("/").pop() ?? null
   transform: translateX(-50%)
   background: rgb(0, 0, 0)
   color: rgba(255, 255, 255)
-  font-size: 10px
+  font-size: 13px
   padding: 3px 8px
   border-radius: 0 0 4px 4px
-  z-index: 1
+  z-index: 9999
   pointer-events: none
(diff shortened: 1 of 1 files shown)


client/src/components/LocalDevBanner.vue
```diff
@@ -0,0 +1,25 @@
...
+  top: 0
+  left: 50%
+  transform: translateX(-50%)
+  background: rgb(0, 0, 0)
+  color: rgba(255, 255, 255)
+  font-size: 10px
+  padding: 3px 8px
+  border-radius: 0 0 4px 4px
+  z-index: 1
+  pointer-events: none
```
gmeben:
```suggestion
  pointer-events: none
  box-shadow: 0 0 1px 1px #fff, 0 0 0 2px #000
```
This will help distinguish the tab from the background on both Light and Dark interfaces.

PR #2241, opened 2026-03-20
feat: improvements to admin interfaces for publications and users.

wreality:


backend/tests/Api/PublicationTest.php
```diff
@@ -445,6 +445,53 @@ public function testQueryFiltersHiddenItems()
         $this->assertCount(2, $json);
     }
 
+    public function testVisibleScopeHidesNonPublicFromRegularUser()
```
gmeben:
```suggestion
    public function testUnassignedUserCannotSeeHiddenPublication()
```

wreality:
@gmeben just a heads up that I noticed I had missed some auth gating on the schema changes.  Its updated now and tests added

backend/graphql/submission.graphql
```diff
@@ -5,10 +5,14 @@ extend type Query {
...
-        @can(ability: "viewAll", query: true)
-        @paginate(defaultCount: 100)
+    submissions(
+        status: [SubmissionStatus!] @scope(name: statusFilter)
+        my_roles: [SubmissionUserRoles!] @scope(name: myRoleFilter)
+        publication: [ID!] @scope(name: publicationFilter)
+    ): [Submission!]
+        @paginate(defaultCount: 100, scopes: ["visible"])
         @orderBy(column: "created_at", direction: DESC)
+        @guard
```
gmeben, line 13: It seems this change now results in the Application Administrator not being able to query for most submissions now. There's only one submission visible on the Submissions and Reviews pages when logged in as this user.

gmeben (changes requested):
I'm getting failing Cypress tests locally under the following specs: 

* copywriting.cy.js (1 failing test)
* reviewspage.cy.js (1 failing test)
* admin/publications.cy.js (3 failing tests)
* admin/user_details.cy.js (2 failing tests)
* admin/users.cy.js (1 failing test)

I'm also encountering that problem I encountered back in March with `schema.graphql` re-rendering single-line syntax descriptions with multiple-line syntax (triple quotes).

backend/graphql/submission.graphql
wreality: reverted the visibility scope and removal of the application admin policy gate, wasn't actually updated in client code, so I think this was just an errant inclusion from earlier work

wreality:
> I'm getting failing Cypress tests locally under the following specs:
> 
> * copywriting.cy.js (1 failing test)
> * reviewspage.cy.js (1 failing test)
> * admin/publications.cy.js (3 failing tests)
> * admin/user_details.cy.js (2 failing tests)
> * admin/users.cy.js (1 failing test)
> 
> I'm also encountering that problem I encountered back in March with `schema.graphql` re-rendering single-line syntax descriptions with multiple-line syntax (triple quotes).

I believe cypress should be passing now.  the schema.graphql problem is likely fixed as well, but it *will* render everything with """ for consistency, which should keep it from conflicting if the schema content is the same. (it doesn't rely on lighthouse's print-schema output for ordering or formatting anymore.

gmeben (changes requested):
I'm encountering this server side error for the `users` operation when visiting the `/admin/users` page which results in a "No data available" message being shown in the table:

```
"extensions": {
    "debugMessage": "Call to undefined method Illuminate\\Database\\Eloquent\\Builder::search() in @scope directive on search argument.",
    "file": "\/app\/backend\/vendor\/nuwave\/lighthouse\/src\/Schema\/Directives\/ScopeDirective.php",
    "line": 39,

```

wreality:
> I'm encountering this server side error for the `users` operation when visiting the `/admin/users` page which results in a "No data available" message being shown in the table:

I missed that the search scope got dropped when master was merged in the other day.  Its fixed.

client/src/components/tables/QueryTable.vue
```diff
@@ -0,0 +1,306 @@
+<template>
+  <q-table
+    v-bind="tableProps"
+    v-model:pagination="paginationModel"
+    flat
+    binary-state-sort
+    table-class="q-table--bordered"
+    table-header-class="bg-accent text-white"
```
gmeben:
```suggestion
```

The accent color in Pilcrow feels more like it should be reserved for decorative or smaller interactive elements, not table headers. There's no precedent for it being used like this so it looks out of place.

client/src/pages/Admin/components/UserSubmissionsNoDataSlot.vue
```diff
@@ -0,0 +1,45 @@
...
+        class="bg-yellow-4 text-black"
+      >
+        <template #avatar>
+          <q-icon name="o_filter_alt" />
+        </template>
+        Head's up: Your filters will not return any results.
+        <span v-if="statusFilter.length === 0"> (no status selected) </span>
+        <span v-if="roleFilter.length === 0"> (no role selected) </span>
+        <template #action>
+          <q-btn label="Reset filters" flat @click="$emit('resetFilters')" />
```
gmeben: Need i18n strings for the text in this file.

client/src/routes/admin/publications.vue
```diff
@@ -0,0 +1,269 @@
...
+  params: { id }
+})
+
+const columns: QueryTableColumn[] = [
+  {
+    name: "name",
+    field: "name",
+    align: "left",
+    sortable: true,
+    label: "Name"
```
gmeben: i18n strings are needed for the labels in this file.

client/src/routes/admin/publications.vue
```diff
@@ -0,0 +1,269 @@
...
+    sortable: true,
+    component: DateTimeCell,
+    label: "Created"
+  },
+  {
+    name: "actions",
+    field: "id",
+    align: "right",
+    label: "Actions"
+  }
```
gmeben:
```suggestion
  }
```

I think the Actions column in this table can be safely removed since it only links to pages (`/publication/#/setup/basic`) that are nested within those that are already linked in the Name column (`/publication/#`). And removing this also eliminates interface problems introduced by having a "Configure" button labeled with text and an icon and only one action available, as opposed to the icon-only labels and multiple possible actions in the Actions columns of submissions tables. 

Removing this also means code around lines 40 and 128 can be removed.

client/src/pages/Admin/components/PublicationsFilterPanel.vue
```diff
@@ -0,0 +1,95 @@
+<template>
+  <q-btn
+    flat
+    dense
+    no-caps
+    icon="filter_list"
+    :label="activeCount > 0 ? `Filter · ${activeCount} active` : 'Filter'"
```
gmeben: This should be in an i18n template string where `${activeCount}` is a property as a parameter.

gmeben (changes requested):

client/src/components/tables/QueryTable.vue
wreality: I can agree with that.  I feel like it needs something to visually separate it though.  How about `bg-grey-2` (obviously without `text-white`)?

gmeben: There would need to be a corresponding dark mode color defined. I believe a reason we didn't use color to differentiate table headings from bodies in submissions tables was to avoid this problem. 

Maybe we use a bold font weight as a less contentious differentiator. I'm kind of surprised Quasar overrides the initial bold font weight of headings with a font weight of only 500. This font weight does not appear differently from normal font weight in Atkinson. It's only subtly more bold in Quasar's default Roboto font.

gmeben: Screenshot of what a table looks like in dark mode with bg-grey-2 applied

wreality:
@gmeben Addressed your notes (apologies re: all the i18n stuff, I overlooked that).
* Refactored how `QueryTable` handles column labels' i18n
* Added docs for `QueryTable` and `Breadcrumbs` to explain conventions and configuration options
* Add dark mode header for `QueryTable`
* Remove actions column in Publications table
* Add vitests for new components

gmeben:
No worries. I've made that same mistake several times, adding temporary labels to get things working and then forgetting because of other things preoccupying.

PR #2257, opened 2026-04-08
chore(cypress): extract cypress into top-level package and upgrade to v15

wreality:
## Summary

Reorganizes cypress tests into their own top-level `cypress/` package so they can be installed and run independently of the Quasar client build, then upgrades cypress to v15.

- **Extract** `client/test/cypress` → `cypress/` with its own `package.json`, `yarn.lock`, eslint config, and `cypress.config.cjs`. Removes cypress deps and eslint overrides from `client/`. Updates root `package.json`, `.lando.extras.yml`, the e2e workflow, and docs to reference the new path.
- **Upgrade** cypress to v15, fix browser launch config, and resolve SSL / node engine errors in the cypress container.
- **CI** add a `lint-cypress` job and a `cypress` path filter so cypress-only changes trigger the right jobs.

## Test plan
- [x] `lando rebuild` brings up the cypress service cleanly
- [ ] `cd cypress && lando cypress run` passes
- [ ] `cd cypress && yarn lint` passes
- [ ] CI `lint-cypress` job runs on cypress-touching PRs
- [ ] e2e workflow (`do-test-e2e`) runs cypress from `cypress/`


gmeben (commented):
I can't get cypress to install or run. `cd cypress && lando cypress run` results in the error: 

> error Command "cypress" not found.

wreality:
Bumped the service to Node 22 / `cypress/base:22.20.0` in 96ac7d2f. To pick up the change locally, you'll likely need to disable and re-enable the cypress extra so the old template isn't stuck in your `.lando.local.yml`:

```sh
lando extras disable cypress
lando extras enable cypress
lando rebuild -y -s cypress
lando cypress run
```

PR #2267, opened 2026-04-22
chore(dx): developer-experience cleanup — test DB isolation, HMR, GraphiQL CSRF, seeder pruning

wreality:
## Summary

A grab bag of developer-experience fixes pulled out of feature branches so they can land independently. None of them change product behavior; they make local development less annoying and remove dead weight from the seeder directory.

### Isolate tests on a separate `laravel_testing` database

`artisan test` (RefreshDatabase) was wiping the developer's local Lando database every time the suite ran. This adds a `testing` DB connection in `config/database.php` pinned to `laravel_testing` (overridable via `DB_TEST_DATABASE`), a symlinked `testing-schema.sql` so Laravel finds the schema dump by connection name, and teaches Lando's database service to create `laravel_testing` and grant the `laravel` user access on build.

The `DB_CONNECTION=testing` override lives on the `backend-test` Lando tooling env only — the test suite itself (phpunit.xml, `.env.testing`) is left at master state so CI still hits the `mysql` connection with the creds `.env.testing` already carries. Dotenv's `safeLoad` doesn't overwrite the process env, so Lando's tooling override wins locally without having to touch anything at the suite level.

`lando artisan test` remains the primary documented command (unchanged for folks used to it). `lando backend-test` is documented as an opt-in for tight iteration loops and agent-driven workflows where clobbering the dev DB is inconvenient. `docs/developers/testing.md` now explains both, including the `--database=testing` flag for targeting the isolated DB with other artisan commands.

Three existing notification tests (`InlineCommentsTest`, `InvitationsTest`, `OverallCommentsTest`) were missing the `RefreshDatabase` trait. On master they worked by accident because tests hit the same `laravel` DB devs had already migrated; on a freshly-created `laravel_testing` they failed with "users table doesn't exist" on the first run. Added the trait.

### Loosen Vite HMR `allowedHosts` to any `*.lndo.site`

Non-default Lando worktrees (e.g. `pilcrow-publication-admin.lndo.site`) were being rejected by HMR. Wildcarding the subdomain lets any worktree connect without hardcoding each one.

### Fix GraphiQL CSRF after auth mutations

The bundled `mll-lab/laravel-graphiql` template bakes `csrf_token()` into a meta tag at render time, so the token goes stale after Laravel rotates it (e.g. on `login`). Every request after that returned 419 until the page was reloaded. Overridden template reads the `XSRF-TOKEN` cookie fresh on each fetch, so devs can keep iterating in GraphiQL after mutating auth state.

Also added a subtle GraphQL-logo icon to the non-default-worktree `LocalDevBanner` that opens `/graphiql` in a new tab, so devs can hop into the IDE without typing the URL.

### Prune orphaned seeders

Production bootstrap already runs via `PermissionSeeder` on the `SchemaLoaded` event, and nothing in the prod path invokes `db:seed`. Cleanup of the dev/Cypress-fixture side:

- Delete `RoleSeeder` (deprecated, zero callers, would crash post-`SchemaLoaded`).
- Delete `UxSeeder` (orphan, manual-only, dormant since 2023).
- Fold `StyleCriteriasSeeder` into `PublicationSeeder` so publication 1's 4 named criteria are attached at creation instead of via a separate order-coupled seeder.
- Trim the 50 filler publications to 3 — Cypress doesn't assert on count and the rest were purely visual noise.

## Test plan

- [x] `lando destroy && lando start` provisions `laravel_testing` and grants `laravel` access.
- [x] On a truly-blank `laravel_testing` (`lando artisan db:wipe --database=testing --force`), `lando backend-test` passes 426/426 on the first run without needing a prior seed.
- [x] `lando backend-test` leaves the dev `laravel` DB untouched (seed dev data, run tests, re-check dev data — `laravel_testing` gets populated instead).
- [x] HMR works on a non-default worktree URL (e.g. `pilcrow-<worktree>.lndo.site`) without Vite rejecting the host.
- [x] GraphiQL: log in via the `login` mutation, then run any authenticated query — it should succeed instead of 419.
- [x] Dev banner (on a non-`pilcrow` worktree dir) shows a GraphQL icon; clicking opens `/graphiql` in a new tab.
- [x] `artisan db:seed` produces 5 publications total (pub 1 with 4 named criteria, pub 2 for rejected-submissions, 3 filler), 6 named users, and the usual submission/comment fixtures.


wreality:
@gmeben This is a relatively small Dx update.  If you have a chance to review, I'd be appreciative.

docs/developers/testing.md
```diff
@@ -24,6 +24,27 @@ To run the PHP unit tests execute the following command **from the `/backend` di
...
+::: tip Isolating tests from the dev database
+`lando artisan test` runs against the same `laravel` database that drives your local UI, so `RefreshDatabase` and `migrate:fresh` will wipe the data you've been working with in the app.
+
+If you'd rather not have the suite clobber that data — useful when iterating tightly (including agent-driven workflows), or when you've spent time setting up specific local state — use the `backend-test` tooling command instead:
+
+```sh
+lando backend-test
+```
+
+It's cwd-independent and points the suite at a dedicated `laravel_testing` database so your dev data stays intact. Note that running PHPUnit directly (e.g. via `lando ssh`) bypasses this override and will still hit the dev database.
```
gmeben:
```suggestion
It works independent of the current working directory and points the suite at a dedicated `laravel_testing` database so your dev data stays intact. Note that running PHPUnit directly (e.g. via `lando ssh`) bypasses this override and will still hit the dev database.
```

I encourage avoiding acronyms for accessibility, especially when the acronyms are less common or unclear.

gmeben (approved):

PR #2284, opened 2026-05-06
fix: serve app banner config from Laravel runtime not build env

wreality:
## Summary
- `APP_BANNER`, `APP_BANNER_CLASS`, `APP_BANNER_LINK` were declared in `quasar.config.ts` `build.env` but never propagated through `docker-bake.hcl` or the `mesh-research/github-actions/build-image` action, so the feature was inert at every deploy
- Move banner config to Laravel (`backend/config/app.php`), serve it to the SPA via `window.__APP_BANNER` from the existing `index.blade.php` injection block (same pattern as `window.__toCdnUrl`)
- Banner is now toggled via FPM container env (Helm values → ConfigMap → env) without rebuilding the client image

## Test plan
- [x] Set `APP_BANNER=Test message` (and optionally `APP_BANNER_CLASS`, `APP_BANNER_LINK`) on the FPM container, reload the SPA, confirm banner renders
- [x] Unset `APP_BANNER`, reload, confirm banner does not render and there are no console errors
- [x] Confirm \`window.__APP_BANNER\` is present in the served HTML when the SPA index is rendered by Laravel
- [x] Verify dismiss-for-a-week behavior still works (no regression in `LocalStorage` flow)

## Local dev note
To test the banner locally, set `APP_BANNER` (and optionally `APP_BANNER_CLASS`, `APP_BANNER_LINK`) in your `.env`. Whenever you change those values, you must:
1. `lando rebuild -s appserver` (or `lando restart`) to refresh the appserver env
2. Hard reload in the browser to refetch the index from Laravel — banner config is injected server-side into `index.blade.php`, so a soft reload of the SPA will not pick up new values

### DIFF
--- a/backend/config/app.php
+++ b/backend/config/app.php
@@ -17,7 +17,7 @@ return [
 
     'name' => env('APP_NAME', 'Pilcrow'),
 
-    'banner' => env('APP_BANNER'),
+    'banner_text' => env('APP_BANNER_TEXT'),
     'banner_class' => env('APP_BANNER_CLASS'),
     'banner_link' => env('APP_BANNER_LINK'),
 
--- a/backend/config/app.php
+++ b/backend/config/app.php
@@ -17,6 +17,10 @@ return [
 
     'name' => env('APP_NAME', 'Pilcrow'),
 
+    'banner' => env('APP_BANNER'),
+    'banner_class' => env('APP_BANNER_CLASS'),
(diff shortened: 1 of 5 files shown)


gmeben (approved):
I suggest changing 'banner' to 'banner_text' just to make references clear that they're referencing the banner's text and not the banner itself, but this is fine to merge.

PR #2286, opened 2026-05-08
fix(api): redact User.email for unauthorized viewers

wreality:
## Summary

Closes the vector where User.email could be read through any GraphQL path that returned a User, including:

- Guest hitting `user(id) { email }` or `userSearch { data { email } }`
- Authed unaffiliated users hitting `user(id) { email }`
- Submitters reading reviewer / review-coordinator emails through their own `submission { reviewers { email } }`

### Approach

- **`@redactIfDenied(ability: ...)` field directive**: returns `null` when the viewer fails the gate instead of throwing, so sibling fields keep resolving.
- **`UserPolicy::viewEmail`**: viewer must be self, an application administrator, or a publication admin / editor for any publication the target belongs to (directly or via one of the target's submissions).
- **`User.email`**: now `String @redactIfDenied(ability: \"viewEmail\")` — every path through related types redacts for unauthorized viewers.
- **`Query.user(id)`**: gated to application administrators via `UserPolicy::view`. `currentUser` remains the path for self.
- **Perf**: viewer's privileged-publication ID lookup is memoized on the model instance, so list resolvers (e.g. `submission.reviewers`) don't re-run the same pivot query for every target user.

### Follow-on fixes in this PR

- **Login / reset-password / oauth-register**: explicitly call `$context->setUser($user)` so the just-authenticated viewer (not the snapshot taken at request start) is used when resolving `email` on the returned User. Adds `LoginMutationTest` regression for the visible-email case.
- **Stable avatars without email**: `User.avatar_color: String! @deprecated` exposes a server-computed color name (JS-parity hash of email, id fallback) so existing avatars stay stable now that emails are redacted. Client `avatarImage` fragment plus `relatedUserFields` / `currentUserFields` / `GET_USER` / `GET_USERS` / `SEARCH_USERS` updated to select it; `AvatarImage.vue` drops its local hash. Field is short-lived — will be removed when user-uploaded avatars land.
- **Schema formatting**: `client/src/graphql/schema.graphql` regenerated to keep triple-quoted descriptions consistent (root-cause codegen fix tracked in `feature/admin-tables`).

### Deliberately not in this PR

- Restructuring `Submission` relations to a restricted `UserLike` interface that omits `email` at the schema level. Discussed and parked — requires touching client fragments / codegen / many consumers and isn't needed to plug the leak.
- Moving `userSearch` under `Publication` and gating it. Was prototyped, reverted as too invasive for this fix.

## Test plan

- [x] `lando backend-test` — full suite passes (includes new `LoginMutationTest`, `UserTest::testAvatarColor*` cases)
- [x] New coverage in `UserPolicyTest` for self / app admin / pub admin / editor / submitter-via-submission-publication / reviewer denied / review coordinator denied / unaffiliated denied / cross-publication editor denied / `view` admin-only
- [x] New coverage in `UserQueryTest`: guest can't call `user(id)`, non-admin can't call `user(id)` (even for self), submitter can't read reviewer/coordinator emails through `submission`, self can read own email via `currentUser`, guest can't read emails via `userSearch`
- [x] `LoginMutationTest`: viewer email is returned (not redacted) on the just-authenticated user
- [x] `UserTest`: `avatar_color` matches legacy client hash for known email; returns one of the known colors
- [x] Memoization test verifies viewer's `publication_user` lookup fires once when policy is checked against many targets
- [x] `lando client-yarn vue-tsc --noEmit` clean
- [x] `lando client-yarn vitest run` — 258 passed
- [x] `lando ssh -s appserver -c \"cd /app/backend && vendor/bin/phpcs\"` clean
- [x] Schema snapshot regenerated; `SchemaSnapshotTest` passes


backend/app/Policies/UserPolicy.php
```diff
@@ -4,13 +4,27 @@
...
 
 class UserPolicy
 {
     use HandlesAuthorization;
 
+    /**
+     * Determine whether the viewer can fetch another user's record via the
+     * top-level user(id) query. Restricted to application administrators.
+     */
+    public function view(User $viewer, User $target): bool
```
gmeben: Isn't this also applied to users viewing their own user records?

backend/app/GraphQL/Directives/RedactIfDeniedDirective.php
```diff
@@ -0,0 +1,75 @@
...
+     */
+    public function __construct(
+        protected Gate $gate
+    ) {
+    }
+
+    /**
+     * Run the resolver only if the current user passes the given ability
+     * against the parent model. When denied (or unauthenticated), return
+     * null instead of throwing — sibling fields keep resolving.
```
gmeben, line 27: After a user logs in, the `email` field is set to null in the `login` mutation's response. 

After logging in while logged in as a different user, the email value of the newly logged in user is returned. 

The `logout` mutation's response also retains the logged out user's email value. 

Is this behavior intentional?

client/src/graphql/schema.graphql
```diff
@@ -1,4 +1,4 @@
-"""Academic profile data for the profile metadata of a user"""
+"Academic profile data for the profile metadata of a user"
```
gmeben: I'm confused. I thought we were proceeding with triple quote syntax.

backend/app/Policies/UserPolicy.php
wreality: It would be, but in the client we only ever perform that using Query.currentUser (which only retrieves the logged in user so no additional gating isn't needed)

backend/app/GraphQL/Directives/RedactIfDeniedDirective.php
wreality: It wasn't intentional.  Added a test and fixed this edge case.

client/src/graphql/schema.graphql
wreality: We are, the fix for that is in admin-tables and hasn't landed on master yet, so isn't incorporated here. Fixed manually for now.

wreality:
Split the avatar work into #2287 to narrow this PR to a single concern (email visibility). The two reverted commits (`d8f48d0a`, `1d3909cf`) are reverted here via `revert: split avatar_color into separate PR (#2287)` — keeping the history visible rather than force-pushing.

**Do not merge before #2287.** Until that lands and this branch rebases onto master, redacted emails collapse the client-side avatar hash to a single color. Backend + vitest specs stay green; the breakage is runtime-only.

Plan: merge #2287 → rebase `fix/email-privacy` onto master → revert commit and the two original avatar commits drop out of the diff → review the narrower diff.

gmeben (approved):

PR #2297, opened 2026-05-20
fix(publication): scope list query to visible publications

wreality:
## Summary
- `GetPublications` returned `This action is unauthorized` whenever the result set contained any private publication the user didn't have a role on. Root cause: `@can(ability: "view", query: true)` iterates every returned `Publication` and throws on the first denied row.
- Replace the `@can` check with a `visibleToCurrentUser` scope applied via `@paginate(scopes: [...])`, filtering at the DB layer instead of throwing.
- Null-guard `PublicationPolicy::view()` so the single-item query handles unauthenticated guests on non-public publications without a null-deref.

## Visibility rules (list query)
- App admin: sees all
- Authenticated user: public publications + any publication they hold a role on
- Guest: public publications only

## Test plan
- [x] `lando backend-test` — 459 passed
- [x] `lando backend-lint` — clean
- [x] `lando client-yarn lint` — clean (1 pre-existing prettier warning, unrelated)
- [x] `lando client-yarn test:unit:ci` — 265 passed
- [x] New tests cover guest / member / non-member visibility

### DIFF
--- a/backend/app/Models/Publication.php
+++ b/backend/app/Models/Publication.php
@@ -65,6 +65,33 @@ class Publication extends BaseModel
     }
 
     /**
+     * Scope to publications the current user is allowed to see.
+     *
+     * Guests see only publicly visible publications. Application
+     * administrators see everything. Other authenticated users see publicly
+     * visible publications plus any private publication they hold a role on.
+     *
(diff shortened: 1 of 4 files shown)


gmeben (approved):
Great catch

PR #2265, opened 2026-04-20
feat: uploadable user avatars with moderation

wreality:
Based on `feature/admin-tables`. Introduces user-uploaded avatars end-to-end plus a light moderation workflow. Storage is abstracted behind `MEDIA_DISK` so a production deploy can point at any S3-compatible bucket without code changes.

## Core feature

- `spatie/laravel-medialibrary` wired into the `User` model with a single-file `avatar` collection. JPEG / PNG / WebP only, 5 MB cap, `thumb` (96×96) and `medium` (256×256) conversions generated on upload.
- GraphQL surface on `User`: `avatar { url, thumb_url, medium_url }` and `avatar_upload_blocked: Boolean!`; `uploadUserAvatar` / `deleteUserAvatar` mutations gated by the existing `@can(update)` on users.
- Uploads are re-encoded through `intervention/image` before being stored — drops EXIF (GPS / device info) and any non-pixel payload a crafted file might carry.
- `UuidPathGenerator` keys media under its stable `media.uuid` rather than the sequential id, so `/storage/…` isn't trivially enumerable.
- `robots.txt` disallows `/storage/`.

## Frontend UI

- `AvatarImage` switches the old colored-bubble PNG placeholder for a jdenticon SVG identicon seeded from the user id; custom palette so it reads on both light and dark surfaces.
- `AvatarUploader` + `AvatarCropDialog` (vue-advanced-cropper) on the account profile page — square crop, client-side resize to 512×512 PNG before upload, checkerboard backdrop so transparent / dark source images stay visible, inline acceptable-use reminder.
- `ReportableAvatar` overlays a small flag/kebab trigger at the bottom-right of any avatar belonging to another user. A `compact` prop is used on 30 px comment-header avatars.
- Report trigger is surfaced in `CommentHeader` and `UserListItem`, self-hides when there's nothing to report (no current user, target has no uploaded avatar, or target == current user).

## Moderation

- `avatar_reports` table + `AvatarReport` model (user, reporter, reason, status, resolver, resolution metadata).
- Mutations: `reportUserAvatar` (any auth'd user, idempotent per pair), `dismissAvatarReport`, `resolveAvatarReportAndRemoveAvatar` with optional `blockFutureUploads` flag.
- `avatar upload revoked` spatie permission — seeded but unassigned; granted to individual users to block future uploads. `setUserAvatarUploadBlocked(userId, blocked)` mutation lets admins grant or revoke independently of any specific report.
[...]

PR #2308, opened 2026-06-11
WIP: feat: uploadable user avatars with moderation (v2)

wreality:
**WIP** — recreated on fresh master, supersedes #2265.

## Scope
- Avatar upload: crop dialog, spatie/medialibrary UUID-path storage, conversions (thumb/medium), queued on a redis worker.
- Moderation: report an avatar, admin review queue with dismiss/remove + pending badge.
- Authorization on Bouncer ABAC (after master #2310 retired Spatie laravel-permission):
  - `moderate avatars` → `GlobalAbility::AvatarModerate`, enforced via a `moderateAvatars` Gate + `@can` directives; app-admin holds it through Bouncer `everything()`.
  - Avatar-upload-block is **moderation state, not a permission**: stored in a new extensible `moderation_flags` JSON column on `users` (`ModerationFlag` enum + `has/set/clear` helpers), mirroring `feature_opt_ins`. New flags = new keys, no new columns.
- Computed GraphQL fields via `User` `@method` accessors; mutations in dedicated resolver classes; client ability checks read `abilities` object (`can('avatar_moderate')`).
- Admin dashboard refactored data-driven via the `useNavigation` composable; users nested under a `users/` section folder so detail routes stay out of the nav tree.
- nginx `/storage` static-serve for dev (`.lando`) and prod (`client/.docker`).

## Status
Merged latest master (Bouncer ABAC #2310 + queue worker). Green: backend 639, client vitest 85 (changed specs), vue-tsc 0, eslint 0, backend-lint + lint:md clean.


backend/app/GraphQL/Mutations/ReportUserAvatar.php
```diff
@@ -0,0 +1,67 @@
...
+        $reportedMedia = $reportedUser->getAvatarMedia();
+        if ($reportedMedia === null) {
+            throw new Error('This user does not have an uploaded avatar to report.');
+        }
+
+        // Idempotent: one pending report per (reporter, reported) pair.
+        $existing = AvatarReport::where('user_id', $reportedUser->id)
+            ->where('reporter_user_id', $reporter->id)
+            ->where('status', AvatarReport::STATUS_PENDING)
+            ->first();
```
wreality: match the reported media as well.  if a new media item needs reporting it should be reocrded.

backend/app/Models/User.php
```diff
@@ -441,4 +451,144 @@ public function getAvatarColorAttribute(): string
...
+
+    /**
+     * Resolve the GraphQL `User.avatar_upload_blocked` field. True when a
+     * moderator has set the avatar-upload-blocked moderation flag on this
+     * user. Uploading is allowed by default; the flag is the exception.
+     */
+    public function getAvatarUploadBlocked(): bool
+    {
+        return $this->hasModerationFlag(ModerationFlag::AvatarUploadBlocked);
+    }
```
wreality, line 501: Do we really need this helper or can we just call hasModerationFlag at the call site ?

backend/app/Models/User.php
wreality, line 1: Should the mioderation flag functionality be refactored into a trait or anogther reusable compoment?

backend/app/GraphQL/Mutations/ReportUserAvatar.php
wreality: Good catch — the idempotency lookup now also keys on `media_id`, so a pending report only short-circuits for the *same* reported image. If the user swaps to a different avatar, that new media is reportable in its own right rather than folded into the stale report. Added `testReportingASwappedAvatarRecordsANewReport` to cover it.

backend/app/Models/User.php
wreality: Kept as a thin accessor because it is the GraphQL seam: the schema binds the field via `@method(name: "getAvatarUploadBlocked")`, and `@method` calls a zero-arg model method — it cannot pass the `ModerationFlag::AvatarUploadBlocked` enum into `hasModerationFlag()` at the call site. So this one-liner is what bridges the schema to the generic flag check. It now delegates to the extracted trait.

backend/app/Models/User.php
wreality: Done — extracted the generic flag logic (`hasModerationFlag` / `setModerationFlag` / `clearModerationFlag`) into `App\Models\Traits\HasModerationFlags`, which `User` now uses. The enum catalog stays in `App\Enums\ModerationFlag`, and consuming models just need the `moderation_flags => array` cast (documented in the trait). Avatar-specific accessors like `getAvatarUploadBlocked()` stay on `User`.

wreality:
## Review — completeness + structural decisions

Overall this is in good shape: strong backend test coverage (39 API tests covering stale resolution, snapshot gating, mime/size, UUID paths), content-based upload validation, EXIF-strip-via-re-encode, private quarantine disk, and ability-gated snapshot serving. The items below are what I'd want resolved before merge, plus a few structural decisions worth recording.

### Structural decisions (agreed)

1. **Moderation state is a subject attribute, not an ability.** `moderation_flags` (mirroring `feature_opt_ins`) is the attribute layer; abilities are decisions derived from it. The decision "may this avatar be uploaded" should be a named, server-owned attribute-based Gate (`uploadAvatar = owner AND !blocked`), consumed by the client as one `can_upload_avatar` boolean — not an imperative `throw` in the resolver re-derived again in client TS.
2. **Abilities are single-subject.** Where an action is reachable by self-service *or* moderation, authorize on a disjunction of named abilities at the call site, never an `|| can(moderate)` folded inside a capability boolean.
3. **Report pins the exact media; snapshot exists while pending.** Pin by `media_id` + durable `reported_media_uuid`; eager copy-on-report; staleness compared by UUID so innocent replacements are never removed; snapshot purged on any resolution. The 90-day post-resolution retention (and its column + scheduled purge command) is **dropped** — no appeals process consumes it, repeat-offender history survives as metadata, and auto-retaining violative content is a liability.
4. **Moderation separates state / queue / history.** Reports are a transient queue; the durable record is the per-user audit history via `owen-it/laravel-auditing` (already a dep). Audit subject = the party an action is about: removals/blocks → reported user; dismissed/upheld → reporter (reporter-reputation signal).
5. **Upload sanitization posture.** Re-encode sanitizes payload/metadata but does **not** contain a malicious decoder — native image-lib patch currency is a security dependency. Reactive moderation (report → remove) is the v1 posture; CSAM perceptual-hash scanning, if required later, belongs on the worker in an upload-private → process → promote-on-pass flow (which also relocates re-encode off the web tier).

### 🔴 Correctness / security — block merge

- [x] **Stale-resolve bulk-close** — the "close other pending reports" block in `ResolveAvatarReport::resolve` runs inside `if ($removeAvatar)` but outside the `$isStale` branch, so resolving a stale report closes everyone else's pending reports as "avatar removed" when nothing was removed (and they may concern the innocent replacement). Move it into the non-stale branch.
- [x] **Uncaught decode failure** — in `UpdateUserAvatar::upload`, `stripMetadata()`'s `$manager->read()` is outside the try/catch, so a corrupt/malicious image that passes mimetype validation but fails GD decode returns a raw 500. Wrap it; return a clean "couldn't process this image" error.
- [x] **Pre-decode dimension/pixel cap** — the 5 MB limit bounds file size, not decoded size; a small file can decode to a gigapixel raster (memory-exhaustion DoS). Check dimensions before full decode.
- [x] **Migration `down()`** — `2026_06_25_..._add_media_and_purge_after` drops `media_id` + `purge_after` but leaks `reported_media_uuid` on rollback.

### 🟠 Authorization refactor

- [x] Define `uploadAvatar` Gate (owner + `!blocked`); gate `uploadUserAvatar` on it; drop the resolver `throw`.
- [x] Add `can_upload_avatar` server field; remove the client `v-if="!avatar_upload_blocked"` re-derivation (`AvatarUploader.vue`).
- [x] Fix `deleteUserAvatar` gate to `owner OR avatar-moderate` (today `@can('update')` only coincidentally covers a future non-admin avatar-moderator).
- [x] Align ability naming with #2327's `admin_*` convention pre-emptively (avoids a merge conflict): the moderate ability → `admin_avatar_moderate` (snake value == field); resolve via Bouncer directly; drop the `moderateAvatars` Gate alias.
- [x] Tests: add "admin clears a *blocked* user"; the existing "admin uploads for another" tests an accidental capability — drop or invert it (upload is self-service only; moderators clear, never replace).

### 🟠 Audit history

- [x] Make `User` Auditable, scoped (`$auditInclude` on moderation attributes + custom events, tagged `moderation`) so it doesn't churn on profile/login changes.
- [x] Emit events: `avatar_removed` / `upload_blocked` / `upload_unblocked` (reported user); `report_dismissed` / `report_upheld` (reporter). Note: `report_upheld` on bulk-closed siblings depends on the stale-bulk-close fix above being correct.
- [x] Move resolution detail (who/when/notes/outcome) off the report into the audit entry; the report keeps minimal queue state.

### 🟡 Frontend

- [x] `avatar-reports.vue`: refetch the paginated list after dismiss/resolve — the actioned row currently lingers in the PENDING table until manual refresh (only `refetchPendingCount()` runs).
- [x] `ReportableAvatar.vue`: the `@keydown.enter/space.prevent` handlers are empty (preventDefault only) — keyboard users can't open the report menu.
- [x] `AvatarUploader.vue`: a blocked user sees the entire uploader `v-if`'d away with no explanation — show a "uploads disabled by a moderator" message.
- [x] `AvatarImage.vue`: `alt`/`aria-label="User Avatar"` is hardcoded English and non-descriptive.
- [x] Resolve-and-remove: add explicit "this permanently deletes the avatar and the evidence snapshot" copy + confirm step (snapshots are purged on resolution).
- [x] Lower priority: `:loading`/disable on resolve buttons (double-submit); refresh the row's `user.avatar` after removal; confirm dialog on self-remove; `AvatarCropDialog.handleSave` fails silently with no blob.

### 🟢 Near-term, separable

- [x] Rate-limit reporting (volume abuse). Targeted bad-faith reporting (an `avatar_report_blocked` flag, same attribute→ability→audit pattern) is a deliberate fast-follow once the report-dismissal history shows it's needed.
- [x] Report idempotency is a read-then-write race — consider a partial unique index on `(user_id, reporter_user_id, media_id)` where pending.
- [x] Runbook note: libwebp/libpng/libjpeg/GD patch currency is a security dependency (re-encode parses hostile input before sanitizing it).
</content>


---

**Progress (pushed to `feature/user-avatars-v2`):** All review items are now addressed across 5 commits (`1683f3a6`, `058bc0a3`, `9dc8e58b`, `d9a2919f`, `24c32500`). Audit history landed as a redesign — reports are now ephemeral (deleted on resolution) and every moderation decision is recorded in the `owen-it/laravel-auditing` log via `User.recordModerationAudit`, surfaced through `User.moderationHistory`. Backend 656 tests green, client vue-tsc + specs green, lint + schema-snapshot clean.

wreality:
Relocating WIP to fork to reduce noise on origin while developing. Continues at wreality/Pilcrow#78. (Will reopen here when ready for review.)

PR #2310, opened 2026-06-16
feat(auth): revamp permissions — RBAC → ABAC (Bouncer), retire Spatie

wreality:
## Summary

Reworks authorization from global, role-presence **RBAC** (spatie/laravel-permission) to **attribute-based access control (ABAC)**. Capabilities are typed abilities, scoping is explicit, conditions are data on the grant, and `spatie/laravel-permission` is removed entirely. The one genuinely global role (application admin) moves to [Bouncer](https://github.com/JosephSilber/bouncer); the scoped publication/submission roles are a code-owned matrix.

Current-state architecture is documented in **[docs/developers/authorization.md](docs/developers/authorization.md)**; design rationale in **[docs/permissions-abac-design.md](docs/permissions-abac-design.md)**. This description covers *what changed*.

## Why

Roles were global in Spatie but most are inherently scoped to a publication or submission. Scoped behavior was hand-rolled on pivots, every capability was a hardcoded role-id list inside policy PHP (edit + redeploy to change), and only one of six roles actually used Spatie. Not maintainable or expandable.

## Two engines, selected by ability *type*

There is no single front door — the ability's type picks the engine, so a Bouncer grant can never short-circuit a scoped check:

- **`App\Auth\GlobalAbility`** (application-wide: `publication.create`, `user.*`) → checked directly against Bouncer at the call site: `$user->can(GlobalAbility::UserViewAny)`. The app admin holds all of them via Bouncer's `everything()`.
- **`App\Auth\ScopedAbility`** (publication/submission) → `App\Auth\ScopedAbilityResolver::allows($user, $ability, $entity)`, which resolves the user's effective scoped roles for the entity (pivots + inherited parent-publication admin) and consults the **code-owned** role→ability matrix. Its `allows()` is typed to `ScopedAbility`, so a global ability is a `TypeError` — it can't be answered here.

`ScopedAbility` is a marker interface implemented by two enums split by model for organization: **`SubmissionAbility`** and **`PublicationAbility`**.

## Roles & the ability matrix

- **Roles are slug-backed enums.** `App\Auth\GlobalRole` (just `application_admin`) and `App\Auth\ScopedRole` (`publication_admin`, `editor`, `review_coordinator`, `reviewer`, `submitter`). `toSlug()` is the backing value; `title()` is `@deprecated` (titles belong in i18n).
- **The scoped role → ability map is code on `ScopedRole`.** Each case lists its grants in shorthand — a bare ability is absolute, an `[ability, PredicateClass]` pair is conditional — normalized to `App\Auth\Grant` (an ability + optional `App\Auth\Predicate`). Roles compose as supersets by spreading. Add a capability = add an entry; live on deploy, no seed.
- **Conditions are data on the grant**, not ability-name hacks or policy branches: e.g. submitter status-edit only while `DRAFT` is `[SubmissionAbility::UpdateStatus, SubmissionIsDraft::class]`. Ownership/relationship one-offs (comment author, `viewEmail` shared-publication) stay in the policy.
- Only `application_admin` is a real Bouncer role (its row + `everything()` grant established by migration/`AbacSeeder`); the scoped roles have no Bouncer rows. `$user->isApplicationAdministrator()` replaces `hasRole(...)`. The dead `hasPublicationRole` / `hasSubmissionRole` (and a latent `wherePivot` bug) are gone.

## Pivot storage: `role` slug added, `role_id` retained (frozen)

- A `role` **slug** column is added to `publication_user` / `submission_user` / `submission_invitations` and backfilled from `role_id`. Authorization keys on the slug; `ScopedRole` is slug-backed, so the pivot value maps straight to a case.
- The legacy integer **`role_id` is kept** — its FK to the spatie roles table is dropped and the column made nullable — as the historical record, so the cutover is recoverable if the backfill goes wrong. New writes set `role` only. **Dropping `role_id` is a deliberate later PR.**

## GraphQL / client

- `PublicationRole` / `SubmissionUserRoles` enums map names → slugs; **enum names are unchanged, so the client contract is preserved** — no client changes. `UserRoles` keeps integer values: it is the `highest_privileged_role` **rank** scale (`GlobalRole::rank()` / `ScopedRole::rank()`), a UI hint, not a stored id.
- `application_admin`'s Bouncer role carries the title, so `currentUser.roles { name }` is unchanged.
- The unused permissions layer is removed: GraphQL `Permission` type, `User.permissions`, `Role.permissions`, the `permission(id)` query.

## List filtering unified with item authz (Tier 1)

Query builders used to reimplement authorization as SQL, parallel to the resolver — a latent security bug (listing what you can't view, or vice versa). They now derive from the **same matrix**:

- `ScopedRole::rolesGranting($ability)` inverts the matrix; `pivot()` / `grantingSlugsFor()` narrow to the role slugs on a pivot.
- `PublicationBuilder::whereCan()` / `SubmissionBuilder::whereCan()` build the WHERE from those slugs, with the same app-admin bypass and parent-publication inheritance the resolver applies. `visible()` delegates to `whereCan(...View)` — fixing a dead submission scope that had no admin bypass and only an existence check.
- `ScopedAbilityListParityTest` asserts the listed set equals the per-item resolver verdict, so the two cannot drift. Conditional abilities (no SQL form for a predicate) throw rather than filter wrongly.

## Removed

`spatie/laravel-permission`, `config/permission.php`, `PermissionSeeder`, the `Role`/`Permission` factories and `app/Models/Role.php`/`Permission.php`. A migration drops the orphaned spatie tables; `InstallationServiceProvider` seeds `AbacSeeder`.

## New / key files

- `app/Auth/ScopedAbilityResolver.php` — scoped pivots → code-matrix bridge
- `app/Auth/ScopedRole.php` — slug enum: the role→ability matrix, `rolesGranting()` / `pivot()`
- `app/Auth/GlobalRole.php` — slug enum for the global role
- `app/Auth/ScopedAbility.php` (interface), `SubmissionAbility.php`, `PublicationAbility.php`, `GlobalAbility.php`
- `app/Auth/Grant.php`, `Predicate.php`, `Predicates/SubmissionIsDraft.php`
- `app/Builders/{Publication,Submission}Builder.php` — `whereCan()`
- `database/seeders/AbacSeeder.php` — seeds the global app-admin Bouncer role
- migrations: `*_create_bouncer_tables`, `*_drop_role_id_foreign_keys_on_pivots`, `*_seed_bouncer_application_admin_role` (ports existing admins), `*_drop_spatie_permission_tables`, `*_add_role_slug_to_pivots`
- `tests/Feature/ScopedAbilityListParityTest.php`

## Testing

- **593 backend tests pass**; `phpcs` and `markdownlint` clean.
- 41 **characterization tests** (`PublicationPolicyTest`, `SubmissionPolicyTest`) were written first to lock pre-existing authorization behavior and stayed green throughout — the migration is behavior-preserving. `ScopedRoleTest` unit-tests the matrix DB-free; `ScopedAbilityListParityTest` guards list/item parity.

How to verify locally:
1. `lando artisan migrate` (bouncer tables, drop spatie FK, seed/port admin, drop spatie tables, add role slug)
2. `lando backend-test`
3. Exercise authz: editor/reviewer/coordinator/submitter on a submission, publication admin, application administrator, plus a guest on a public vs hidden publication.

## Follow-ups (non-blocking)

- Drop the retained `role_id` column once the slug column is proven in production.
- List-filtering **Tier 2** (`Predicate::scope()` so conditional abilities are list-filterable) and **Tier 3** (fold `myRole` / `myRoleFilter` / `managedPublicationSubmissions` fully onto `whereCan`).
- Per-entity capability flags to the client (replace role-string gating in `.vue`).
- Regenerate `mysql-schema.sql` to bake in `bouncer_*` and the `role` column.
- `UserPolicy::update` TODO for "update users within own publication" — now a one-line ability if desired.


backend/app/Auth/Roles/GlobalRole.php
wreality, line 1: If we implement this as a slug backed enum with the interface:
GlobalRole::ApplicationAdministrator->toSlug() (alias for value/slug)
GlobalRole::ApplicationAdministrator->title() (mark @deprecated bc this should be handled in i18n from the slug

backend/app/Auth/Abilities/GlobalAbility.php
```diff
@@ -0,0 +1,29 @@
...
+ * The backing value is the Bouncer ability name.
+ */
+enum GlobalAbility: string
+{
+    case PublicationCreate = 'publication.create';
+
+    case UserView = 'user.view';
+    case UserViewAny = 'user.view-any';
+    case UserUpdate = 'user.update';
+    case UserManageBeta = 'user.manage-beta';
```
wreality, line 28: These not being namespaced in any way feels like a smell to me, but we probably don't need to add another entire type to accomplish UserAbility::view.  This may be overcomplicating, but suggestions are welcome.

backend/app/Auth/ScopedAbilityResolver.php
```diff
@@ -0,0 +1,139 @@
...
+            );
+        }
+
+        // Dedupe by backing value (enum instances are not array_unique-able).
+        $unique = [];
+        foreach ($roles as $role) {
+            $unique[$role->pivotValue()] = $role;
+        }
+
+        return array_values($unique);
```
wreality: Where would duplication come from in this instance ?

backend/app/Auth/ScopedRole.php
```diff
@@ -0,0 +1,164 @@
...
+ * global roles (App\Auth\GlobalRole, e.g. application_admin), which is intentionally
+ * NOT a case here.
+ */
+enum ScopedRole: int
+{
+    case PublicationAdmin = 2;
+    case Editor = 3;
+    case ReviewCoordinator = 4;
+    case Reviewer = 5;
+    case Submitter = 6;
```
wreality: the enum should be backed by slugs, Ids are legacy ids

backend/app/Models/Publication.php
```diff
@@ -75,7 +76,7 @@ public function users(): BelongsToMany
     public function publicationAdmins(): BelongsToMany
     {
         return $this->users()
-            ->withPivotValue('role_id', Role::PUBLICATION_ADMINISTRATOR_ROLE_ID);
```
wreality, line 78: We're already touching these now that we're using scoped ability, so we should make the change to pivoting on slugs rather than ids after all.

backend/app/Auth/Roles/ScopedRole.php
```diff
@@ -0,0 +1,279 @@
...
+        return $this->legacyId();
+    }
+
+    /**
+     * Human-readable title (the Bouncer-era display name), used where a scoped
+     * role is surfaced by title rather than id.
+     *
+     * @return string
+     */
+    public function title(): string
```
wreality, line 292: This should be marked @deprecated, display strings should be handled by i18n

backend/app/Models/Submission.php
```diff
@@ -133,7 +140,8 @@ public function submissionAssignments(): HasMany
     public function reviewers(): BelongsToMany
     {
         return $this->users()
-            ->withPivotValue('role_id', Role::REVIEWER_ROLE_ID);
+            ->withPivotValue('role', ScopedRole::Reviewer->toSlug())
+            ->withPivotValue('role_id', ScopedRole::Reviewer->legacyId());
```
wreality, line 147: Should these be using withPivotValue for role_id?  We're just keeping that column updated to preserve our rollback ability.

backend/app/Providers/InstallationServiceProvider.php
```diff
@@ -17,9 +18,20 @@ public function boot(): void
...
+     * @param \Illuminate\Database\Events\SchemaLoaded $event
+     * @return void
+     */
     public function addDefaultRolesWhenSchemaLoaded(SchemaLoaded $event)
     {
-        $seeder = new PermissionSeeder();
-        $seeder->run();
+        if (Schema::hasTable('bouncer_roles')) {
+            (new AbacSeeder())->run();
+        }
```
wreality, line 35: Do we really need to seed the bouncer roles ?  Or does bouncer's Upsert's handle this for us ?

backend/app/Auth/Roles/ScopedRole.php
wreality: Done in a05c898f — added the same `@deprecated` (i18n) docblock as `GlobalRole::title()`.

backend/app/Models/Submission.php
wreality: Yes — these already dual-write `role_id` via `->withPivotValue('role_id', ScopedRole::X->legacyId())` (line 147 here, plus `reviewCoordinators()`/`submitters()` and `Publication::publicationAdmins()`/`editors()`). The slug `role` column is the source of truth; `role_id` is kept in sync purely to preserve rollback ability, exactly as you describe. No change needed.

backend/app/Providers/InstallationServiceProvider.php
wreality: Yes, we need it — but not for the role *row*. Bouncer lazily creates a role row on first `$user->assign('application_admin')`, so that part it would handle. What Bouncer never creates implicitly is the **ability grant**: `Bouncer::allow(...)->everything()`. Without seeding that wildcard, an app-admin would hold the role but **zero abilities** — every `$user->can()` returns false. So the seed establishes what the role can *do*, not that it exists.

It is idempotent: `role()->firstOrCreate(...)` + `allow()->everything()` are no-ops once present (Bouncer skips an ability it already granted).

Re redundancy: the same role+grant is established in three places by design — the `seed_bouncer_application_admin_role` migration (real deploys, also ports existing admins), `AbacSeeder` (fresh `db:seed` + test bootstrap), and this `SchemaLoaded` hook (when the schema dump is loaded before that migration runs). Happy to consolidate if you would prefer a single path.

backend/app/Auth/Roles/ScopedRole.php
```diff
@@ -0,0 +1,281 @@
...
+     *
+     * Tier 1 list-filtering only supports unconditional abilities: a predicate
+     * lives in PHP and has no SQL form, so a conditional grant for the requested
+     * ability throws rather than silently filtering incorrectly.
+     *
+     * @param \App\Auth\Abilities\ScopedAbility $ability
+     * @return array<int, \App\Auth\Roles\ScopedRole>
+     * @throws \LogicException if any role grants the ability conditionally
+     */
+    public static function rolesGranting(ScopedAbility $ability): array
```
wreality, line 86: This should be aggressively cacheable.

backend/app/Auth/Roles/ScopedRole.php
```diff
@@ -0,0 +1,281 @@
...
+     * grant) or an [ScopedAbility, Predicate class-string] pair (conditional grant).
+     *
+     * Roles compose as supersets: each spreads the one below it, so "everything
+     * role B has, plus X" is explicit. The submitter is not in the coordinator
+     * chain — it extends the reviewer with title / submitter edits and a
+     * DRAFT-only status change (a conditional grant).
+     *
+     * @return array<int, \App\Auth\Abilities\ScopedAbility|array{0: \App\Auth\Abilities\ScopedAbility, 1: class-string<\App\Auth\Grants\Predicate>}>
+     */
+    private function grantDefinitions(): array
```
wreality, line 161: This should also be cacheable.

backend/app/Auth/Roles/ScopedRole.php
wreality: Done in 0455bc12 — memoized per process, keyed by ability class+case name. It's a pure function over the code-constant matrix, so the cache lives for the process lifetime; conditional abilities throw before the cache write so they're never memoized.

backend/app/Auth/Roles/ScopedRole.php
wreality: Done in 0455bc12 — memoized per process, keyed by ability class+case name. It's a pure function over the code-constant matrix, so the cache lives for the process lifetime; conditional abilities throw before the cache write so they're never memoized.

PR #2319, opened 2026-06-18
feat(backend): add database/file backups via spatie/laravel-backup

wreality:
## Summary

Wires up [spatie/laravel-backup](https://spatie.be/docs/laravel-backup) so the application defines **what** to back up — the database, and optionally uploaded files — using the standard `backup:run` command. No Pilcrow-specific wrapper. **Where** the archive lands (S3, Spaces, etc.) is a deployment concern; the app just defaults to a working local backup and lets the deployment point `BACKUP_DISKS` at whatever disk it configures.

An earlier revision added a custom `app:backup` command and a dedicated backup S3 disk; both are dropped — everything the command did is reachable through spatie's config + `backup:run` flags, and the destination belongs to the deployment.

## Changes

- **Deps:** `spatie/laravel-backup` + `league/flysystem-aws-s3-v3`. The repo's pre-existing `s3` disk had no adapter installed; adding it fixes that and lets deployments use S3-compatible backup destinations without a code change.
- **`config/backup.php`:** the backup *source* — databases (always) and file paths (`BACKUP_FILE_INCLUDE`, opt-in via `backup:run` without `--only-db`). `BACKUP_DISKS` selects the destination disk(s), default `local`. Failure/unhealthy notifications email when `BACKUP_NOTIFICATION_EMAIL` is set; default backup name falls back to `pilcrow-backup`. Automatic cleanup is **off**.
- **`config/database.php`:** non-locking mysqldump (`useSingleTransaction`) for a consistent snapshot that doesn't block the live app.
- **Docs:** new _Installation → Recipes → Backups_ page — local backup, off-host destinations as a deployment concern, scheduling (cron / k8s CronJob / pre-migration), restore, cleanup-is-off, and a volume-mount escape hatch for fully custom config.

## Usage

```sh
php artisan backup:run --only-db                 # database only -> local disk (default)
php artisan backup:run                            # + files (BACKUP_FILE_INCLUDE)
php artisan backup:run --only-to-disk=offsite     # a disk the deployment configured
# BACKUP_DISKS=local,offsite                       # write to multiple disks every run
```

## Testing

- `php artisan list` shows the spatie `backup:*` commands; config loads clean; disks resolve to `local,public,s3`.
- Verified end-to-end against Lando (PHP 8.4): DB zip written to `local`, exit 0.
- `markdownlint` clean on the new recipe.

## Notes / follow-ups

- The backup **destination** (off-host disk + credentials) and **scheduling** live with the deployment, not the app. Give the destination its own bucket/credentials — don't reuse the app's upload or SES mail credentials.
- Cleanup is **off by default** — `backup:run` never deletes and `backup:clean` is not scheduled, so backups accumulate until retention is managed. Recommend a storage-side lifecycle rule; opt into in-app cleanup by scheduling `backup:clean` and tuning the `cleanup` strategy.


backend/config/backup.php
```diff
@@ -0,0 +1,272 @@
...
+        ],
+    ],
+
+    /*
+     * Automatic cleanup is OFF by default: nothing here runs on its own.
+     * `backup:run` never deletes — old backups are only removed when you run
+     * `backup:clean`, which Pilcrow does not schedule. Leave it that way and
+     * manage retention at the storage layer (e.g. an S3 lifecycle rule), or
+     * opt in by scheduling `backup:clean` and overriding the strategy below.
+     * The values are spatie's defaults, kept so `backup:clean` works if run.
```
wreality: We don't need this much detail here. We can just say this is the default for when the user runs backup:clean

docs/install/recipes/backups.md
```diff
@@ -0,0 +1,106 @@
...
+```
+
+Set `BACKUP_DISKS=local,offsite` to write to several disks every run. Give the backup destination its own bucket and scoped credentials — don't reuse the application's upload or mail (SES) credentials.
+
+::: warning Cleanup is off by default
+Pilcrow does **not** delete old backups automatically. `backup:run` never removes anything, and `backup:clean` (spatie's cleanup command) is not scheduled — so backups accumulate until you manage retention yourself. The recommended approach is a storage-side lifecycle rule (e.g. an S3 / Spaces expiration policy): it's enforced by the storage layer and survives if the app is down.
+
+To opt into in-app cleanup instead, schedule `php artisan backup:clean` on your platform and tune the retention numbers under the `cleanup` key in `config/backup.php` (override it via env or a [volume-mounted config](#fully-custom-configuration-volume-mount)).
+:::
+
```
wreality: Spatie offers a cleanup command, but users should be aware that managing backups is their responsibility. While Spatie provides the backup configuration for the default cleanup, any automation of the cleanup process should include a plan for handling the resulting files.

backend/config/backup.php
wreality: Trimmed to a one-line note: "Default strategy used when the user runs \`backup:clean\`." (33742bc)

docs/install/recipes/backups.md
wreality: Reframed the callout to "Managing backups is your responsibility" — notes that spatie provides the cleanup command + default strategy but nothing schedules it, and that any cleanup automation must account for the resulting files. (33742bc)

PR #2327, opened 2026-06-25
feat(auth): gate the client on viewer abilities instead of roles

wreality:
## Context

PR #2310 revamped authorization RBAC → ABAC (Bouncer) and exposed per-type `abilities` over GraphQL (`User`/`Publication`/`Submission`), but deliberately deferred the client migration. The UI still gated on roles (`effective_role`, `my_role`, `highest_privileged_role`, `roles[].name == "Application Administrator"`), which can drift from real authorization. This PR moves the client's **authorization gates** onto the server-resolved ability flags so they can't.

## Backend — fill the two gaps with no existing ability

- **`GlobalAbility::AccessAdmin`** (`admin.access`) — gate for the admin area. App admin holds it through Bouncer's `everything()` wildcard; surfaced as `UserAbilities.access_admin`.
- **`SubmissionAbility::Export`** (`submission.export`) — granted to `ReviewCoordinator` + `Submitter` (Editor/PublicationAdmin inherit via the superset chain; Reviewer excluded; app admin short-circuits in the resolver). Surfaced as `SubmissionAbilities.export`.
- `AbilitiesTest` extended to cover both flags.

## Client

- Query `abilities` on `currentUser`, submission, and publication operations; regenerated the typed schema snapshot + operations.
- `useCurrentUser`: dropped the role helpers and `checkRole`; added `can(ability)` + `canAccessAdmin` reading the global flags.
- Gates migrated to abilities:

| File | Was | Now |
|---|---|---|
| `AppHeader` | `isAppAdmin` | `canAccessAdmin` |
| router guards | `highest_privileged_role`, `isPubAdminOrEditor`, `user.roles` | `abilities.access_admin`, `submission.abilities.view`/`export`, `publication.abilities.view` |
| `guiElements` | `isReviewer` / role bundle | `!update_status` / `!export` |
| `PublicationHomePage`, `SetupLayout` | `effective_role == publication_admin` | `publication.abilities.update` |
| `AssignedSubmissionUsers` | `effective_role == review_coordinator` | `container.abilities.invite` |

## Deliberately retained (role-as-identity, not authz)

- Dashboard submission **bucketing**, `highest_privileged_role` table props, and role **label** columns.
- **Draft/Preview** guards and the **reviewer-in-nonreviewable-state** denial — no ability distinguishes a reviewer (who also holds `view`) from the parties those flows admit; these are routing nuances, not authorization.

## Verification

- Backend: 610 passed (full suite) + lint clean.
- Client: 506 passed (full unit suite), `vue-tsc` clean, eslint 0 errors.

https://claude.ai/code/session_012EnAtELfBZdwdtQvRuG4uj


backend/app/Auth/Abilities/GlobalAbility.php
```diff
@@ -16,14 +16,20 @@
...
     case PublicationCreate = 'publication.create';
 
-    case UserView = 'user.view';
-    case UserViewAny = 'user.view-any';
-    case UserUpdate = 'user.update';
-    case UserManageBeta = 'user.manage-beta';
+    case AdminUserView = 'user.view';
+    case AdminUserViewAny = 'user.view-any';
+    case AdminUserUpdate = 'user.update';
+    case AdminUserManageBeta = 'user.manage-beta';
```
wreality: Where are these slugs used, and can we instead use them to represent the field name. None of these should be implemented yet other than the eveything() admin role,, so changing slug names shouldn't really cause a problem.

client/src/components/AppHeader.vue
```diff
@@ -186,7 +186,7 @@ interface Props {
 }
 defineProps<Props>()
 
-const { currentUser, isAppAdmin } = useCurrentUser()
+const { currentUser, canAccessAdmin } = useCurrentUser()
```
wreality, line 189: is this better as individual "canAccessAdmin" methods/computeds or would can('access_admin') be a more useful structure ?

client/src/use/guiElements.ts
```diff
@@ -232,32 +232,23 @@ export function useStatusChangeControls(
...
       return true
     }
-    return !(
-      isAppAdmin.value ||
-      isPublicationAdmin(submission.value.publication) ||
-      isEditor(submission.value.publication) ||
-      isReviewCoordinator(submission.value) ||
-      isSubmitter(submission.value)
-    )
+    return !submission.value.abilities?.update_title
```
wreality: This seems fragile given update_title's constraints may change.  Is there a better way to handle this client policy/gate?

backend/app/Auth/Abilities/GlobalAbility.php
wreality: Slugs are the Bouncer ability identifier resolved via `$user->can(GlobalAbility::X)`. Only consumer today is the app-admin `everything()` wildcard — nothing grants them explicitly (confirmed no persisted/literal use beyond unrelated `publication.created` telemetry). Per your point, since these are new/unseeded, renaming is safe: backing values now equal the generated field name (`admin_user_view_any`, etc.), so the Bouncer ability and the `UserAbilities` field are one identifier with no dotted/snake split. `SubmissionAbility` keeps its dotted values (pinned by pivot data); doc notes the distinction. e6d2b0a7

client/src/components/AppHeader.vue
wreality: Kept the named computed but routed it through the generic helper: `canAccessAdmin = computed(() => can("admin_area"))`. Rationale: `admin_area` is not a normal capability but the server-computed *union* of `admin_*` abilities, so a named gate keeps that meaning in one place. Granular admin controls still use the generic form, e.g. `can("admin_user_update")`. So both structures coexist by design. e6d2b0a7

client/src/use/guiElements.ts
wreality: Agreed it was fragile. Switched the export gate to key on `view` instead of the `update_title` proxy — export renders already-viewable content, so "can view + exportable state" is the stable, correct policy. Applied in both `useSubmissionExport` and the route guard so they agree; `isDisabledByRole` → `isDisabledByAccess`. `update_title` was then fully dead on the client and is dropped from the queries. e6d2b0a7

wreality:
Relocating WIP to fork to reduce noise on origin while developing. Continues at wreality/Pilcrow#77. (Will reopen here when ready for review.)

Issue #77, opened 2021-01-07
Utility of document oriented database


wreality:
Previous Card: https://trello.com/c/ot6zelSo/48-determine-development-stack

---

**Problem/Why**

The development team isn't entirely sure where a document store might be useful to the project, but it seems worth investigating.

---

**Goal/What**

Produce research and some opinions to report to the team so that we can make a decision regarding the utility/benefit of adding a document store to the project.

---

**Solution/How
Added notes on research to the CCR architecture document:
https://docs.google.com/document/d/15qTFcvSox0ZxL1gI1V98RxRcqIT2sAob17i7amd-42I/edit#heading=h.i5ej7tig9g64
--- 
### To Do 

- [x] Research usage cases for document stores
- [x] Report that research out to the team for discussion

### Review Checklist 
 - [x] Presentation was helpful to the team and answered the needed questions.

> Trello Source: https://trello.com/c/fq0UNi6X

sattler9:
Will report out at Monday, April 9th during team time

amareshjoshi:
Added notes on research to the CCR architecture document:
(https://docs.google.com/document/d/15qTFcvSox0ZxL1gI1V98RxRcqIT2sAob17i7amd-42I/edit#heading=h.i5ej7tig9g64)

PR #2339, opened 2026-07-10
refactor(client): migrate submission mutations to typed documents

wreality:
## What

Peels the typed-document precursor out of the client ability-gating branch (#77) so it can land independently.

1. **Migrate submission writes off the `updateSubmission` god-mutation** — each client mutation wrapper now points at its intent-shaped replacement (title/content, status, roster updates, comment ops). Operation names and selection sets unchanged; callers untouched.
2. **Retype the 13 migrated submission/comment mutations as `graphql()` documents** — off the untyped `gql` exports into a typed section at the top of `mutations.ts`. Consumers import the generated `*Document` types, so variables and results are type-checked (surfaced and fixed a string-vs-SubmissionStatus gap). Legacy `gql` exports migrate into the typed section as their consumers convert.
3. **Merge the comment `*Reply` mutation documents into their base operations** — the update pairs had identical variables and the create pairs differ only by the optional `reply_to_id`/`parent_id`, which are omitted from the input unless replying (8 documents → 4).
4. **Colocate the user-list fragment on UserListItem** — the component owns a `userListItem` fragment describing exactly what it renders; roster mutations select `...userListItem` instead of a generic shared user fragment.
5. **Share the user-list unassign handler** — AssignedPublicationUsers/AssignedSubmissionUsers carried byte-identical `handleUserListClick` implementations; extracted as `useUnassignUser` in `src/use/userList.ts`.
6. **Keep the GetSubmission refetch on title edits (with a comment saying why)** — the normalized cache covers the title itself, but the audits list on GetSubmission gains a new server-side row per title change that the mutation payload can't supply; the refetch is what refreshes the audit trail (briefly dropped mid-branch, restored in 2f1e6987).
7. **Keep roster mutation writes cache-complete** — the roster mutations select `display_label` alongside `...userListItem` because the watched GetSubmission/GetSubmissionReview queries read it on the same fields; omitting it left the cache diff incomplete for newly-assigned users and forced a full refetch (b6abdf41).
8. **Fix a fallow false positive that gates any PR touching generated-graphql imports** — the codegen output under `src/graphql/generated/` is gitignored, so fallow never indexes it and misreads those `src/*` alias imports as a bare unlisted `src` package (this PR's first CI run and #77 both tripped over it). Ignore the phantom package in `.fallowrc.jsonc`; a genuinely wrong `src/` import still fails vue-tsc.

Documents were briefly colocated in their consuming components mid-branch; that was walked back (58ff6143) — consumers import the generated `*Document` either way, so component-level colocation added file bloat without the fragment/component restructuring that would make it pay. That restructuring is a separate effort; `userListItem` (point 4) is the one piece of it that already earns its keep.

## Why now

Ability gating converts pages to typed colocated documents page-by-page; this is the shared groundwork that every later page conversion builds on. Splitting it keeps the gating PR reviewable on its semantics instead of this mechanical churn.

## Verification

- vue-tsc clean, eslint 0 errors, full vitest suite 506 passing (re-run after each follow-up commit).
- Fallow audit: 0 introduced findings (the unassign-handler extraction clears the duplication the fragment retyping would otherwise introduce).


wreality:
Recreated as a same-repo PR from the branch pushed to this repo (cross-fork was unintended).

